PR #23, opened 2017-09-29
Future proofing and updated journal number

Samwalton9:
Just a minor text update to the front page!


Samwalton9:
Forget this, I didn't realise the previous number was unique periodicals.

PR #35, opened 2018-03-06
Added Admin link, visible only to staff (T169974)

Samwalton9:
Tested with user, coordinator, and staff accounts.

### DIFF
--- a/TWLight/templates/base.html
+++ b/TWLight/templates/base.html
@@ -122,16 +122,16 @@
 				</a>
 			</li>
 			
-		{% if user.is_staff %}
-			<li role="presentation" class="mobile-menu">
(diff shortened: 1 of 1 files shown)


jsnshrmn:
Did you get happy output from:
```
sudo su www /var/www/html/TWLight/bin/virtualenv_test.sh
```

jsnshrmn:
I think you want to nest your is_staff check inside of your is_authenticated check. That way we're not executing both of those evaluations on anonymous users.

Samwalton9:
Nested staff check, and re-ran tests with different account levels + ran test suite with no errors (except for the IOError I sent in Slack)

PR #39, opened 2018-03-07
Application field for having already signed up (T172095)

Samwalton9:
https://phabricator.wikimedia.org/T172095

Functionality seems to work correctly - tested various combinations of link and requirement both in applying and in the admin interface. Tests ran with no problems, though I did have to update the application tests to include already_signed_up first, which is the part I'm least confident about.


Samwalton9:
Oh, this accidentally included a previous PR, hang on.

PR #43, opened 2018-03-08
Fixed median days bug for partners with imported applications (T177200)

Samwalton9:
Imported applications have days_open = None which was breaking the median days statistic on partner pages. Excluding None values from the queryset fixes this.

Tests ran with no issues.


Samwalton9:
Just realised that elsewhere (dashboard) we `.exclude(imported=True)`. Is that preferable?

PR #44, opened 2018-03-12
Median days stat improvements

Samwalton9:
Fixed median days showing 0 for imported applications (T177200). Also fixed it including apps that weren't yet closed, and now displays a dash if the partner has no closed, not imported, applications.

### DIFF
--- a/TWLight/resources/views.py
+++ b/TWLight/resources/views.py
@@ -81,7 +81,7 @@ class PartnersDetailView(DetailView):
         if len(partner_app_time) > 0:
             context['median_days'] = get_median(list(partner_app_time))
         else:
-            context['median_days'] = "-"
+            context['median_days'] = None
 
         context['app_distribution_data'] = get_application_status_data(
                 Application.objects.filter(partner=partner)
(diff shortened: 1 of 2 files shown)


TWLight/resources/views.py
```diff
@@ -74,10 +74,14 @@ def get_context_data(self, **kwargs):
...
+        application_end_states = [Application.APPROVED, Application.NOT_APPROVED, Application.SENT]
         partner_app_time = Application.objects.filter(
-            partner=partner).values_list('days_open', flat=True)
+            partner=partner, status__in=application_end_states).exclude(imported=True).values_list('days_open', flat=True)
 
-        context['median_days'] = get_median(list(partner_app_time))
+        if len(partner_app_time) > 0:
+            context['median_days'] = get_median(list(partner_app_time))
+        else:
+            context['median_days'] = "-"
```
jsnshrmn: This is setting this variable to maybe be an int or maybe be a string. The ideal thing to do here would be to set the value to None, rather than setting it to a string. Then the template can have a conditional for if context['median_day'], because None tests false. You could put your dash in the template on that check, or just let it be blank.

Samwalton9:
Updated per your advice :)

PR #51, opened 2018-03-26
Added some alt text to images without any

Samwalton9:
Benefits and Partners images are decorative but for accessibility reasons should be given empty alt attributes.

### DIFF
--- a/TWLight/templates/home.html
+++ b/TWLight/templates/home.html
@@ -30,7 +30,7 @@
   <div class="row">
 	<div class="col-md-6">
                 {% comment %} Translators: On the main page of the website (https://wikipedialibrary.wmflabs.org/), the 'Benefits' section describes why editors would want to use the project. This is the header for that section.{% endcomment %}
-		<h2><img src="/static/img/benefits.PNG" class="icons">{% trans "Benefits" %}</h2>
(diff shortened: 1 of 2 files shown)


TWLight/templates/home.html
```diff
@@ -30,7 +30,7 @@
   <div class="row">
 	<div class="col-md-6">
                 {% comment %} Translators: On the main page of the website (https://wikipedialibrary.wmflabs.org/), the 'Benefits' section describes why editors would want to use the project. This is the header for that section.{% endcomment %}
-		<h2><img src="/static/img/benefits.PNG" class="icons">{% trans "Benefits" %}</h2>
+		<h2><img src="/static/img/benefits.PNG" class="icons" alt="">{% trans "Benefits" %}</h2>
```
jsnshrmn: adding empty alt attributes is worse than leaving them off.

Samwalton9:
Are you sure? I was following https://www.w3.org/WAI/tutorials/images/decorative/ which says that assistive software will ignore images with empty alt attributes.

jsnshrmn:
I stand corrected: either the guidance has changed, or I've been doing it exactly backwards for years!

PR #55, opened 2018-03-28
Apply button changes to Renew if user already has a sent application (T174942)

Samwalton9:
This is only part of the task at https://phabricator.wikimedia.org/T174942 (the renewal part).

If the user has an open application I'm not sure what the behaviour should be - changing the Apply button to be deactivated is probably the right thing to do, but there should really be a link to their existing application somewhere.


Samwalton9:
Just noticed a bug, hold the phone.

PR #57, opened 2018-03-30
Added specific error message in the case that a coordinator doesn't select any applications to batch edit (T174680)

Samwalton9:
No errors raised, except for the part I removed. Not sure if removing was the right option but it seemed sensible to me given that no error is raised when no applications are selected now; users just reload the page and receive an info message.

### DIFF
--- a/TWLight/applications/tests.py
+++ b/TWLight/applications/tests.py
@@ -2193,10 +2193,6 @@ class BatchEditTest(TestCase):
         response = self.client.post(self.url, data={}, follow=True)
         self.assertEqual(response.status_code, 400)
 
-        # Missing the 'applications' parameter: bad.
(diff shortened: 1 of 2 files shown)


jsnshrmn:
This looks very good. My only suggestion: set the message to warning or error instead of info since the workflow can't proceed unless they make a different choice.

Samwalton9:
Should do the same for https://wikipedialibrary.wmflabs.org/applications/request/

Samwalton9:
Updated INFO messages to WARNING where appropriate.

jsnshrmn:
This PR looks hinky. It should show 4 commits and 2 files changed, but instead it shows 48 commits and 36 files changed.  See the straight up comparison between the two branches:
https://github.com/WikipediaLibrary/TWLight/compare/master...Samwalton9:T174680

PR #71, opened 2018-04-16
Fix 'Send data to partners' page having variables as field labels instead of human-readable labels (T191886)

Samwalton9:
Tests ran OK.

Not sure if this was the best way to do this - we seem to be building this form manually so this feels hacky, but it works.


jsnshrmn:
Yeah, applications have the ugliest code in the whole platform. This is the most straightforward way I see to do it without reworking the applications app entirely.

PR #74, opened 2018-04-16
Fix Apply button disappearing if partner waitlisted (T192290)

Samwalton9:
In https://github.com/WikipediaLibrary/TWLight/pull/61 I accidentally removed the apply button from partner pages if they were waitlisted. Now all waitlisting does to the page is display a notice above the usual button behaviour.

Tests ran OK.


jsnshrmn:
We both missed that.

PR #76, opened 2018-04-17
Add partner location field to resources (T191784)

Samwalton9:


Samwalton9:
Need to update this to allow for translation. Maybe there's a django package that has pre-translated country names?

Samwalton9:
Django-countries seems to work well for what we need - gives staff a drop-down for country names in the admin interface, and translates automatically based on user interface language.

jsnshrmn:
I've resolved the base settings conflict, but now we've got a post-upgrade migration conflict. I had to manually trigger a rebuild in travis, so I'll go do that for the other PRs as well.

Samwalton9:
Fixed the migrations conflict.

PR #80, opened 2018-04-23
Update last_updated when users edit their profile (T192764)

Samwalton9:
I made this change primarily by looking at similar code elsewhere in the site, so this might not be the ideal implementation, but works as far as my testing can tell.

### DIFF
--- a/TWLight/users/views.py
+++ b/TWLight/users/views.py
@@ -1,3 +1,5 @@
+import datetime
+
 from crispy_forms.helper import FormHelper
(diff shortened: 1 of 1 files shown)


Samwalton9:
This date should probably also update if the user's info updates from Wikipedia, but I couldn't figure out where in the code that happens.

Samwalton9:
We could just use last login since info is updated every time they log in.

Samwalton9:
We should just drop the field altogether.

PR #83, opened 2018-04-25
Add alt text to partner logos

Samwalton9:


jsnshrmn:
our alt text should probably be marked as translatable.

PR #84, opened 2018-04-25
Dashboard revamp

Samwalton9:
We still need to rework pageviews, but I'm not confident about how best to do that, so this is a PR for the other work I did to the page.


Samwalton9:
I'm getting the feeling something went wrong here (again). Let me know if this isn't going to work and I'll try to fix it!

jsnshrmn:
Yeah, I'm kind of scared to merge this in, but I know it has good work in it. If you still have a local copy of this branch, maybe you could just push it to a new remote and do another pr?

jsnshrmn:
I just looked at your fork, and diff looks right there. I'm killing this pr and creating a new one from the same fork.

PR #91, opened 2018-05-21
Assorted changes to tidy up the user page

Samwalton9:
Not sure if the way I linked the contributions message is the best way to do that.

### DIFF
--- a/TWLight/users/views.py
+++ b/TWLight/users/views.py
@@ -92,8 +92,11 @@ class EditorDetailView(CoordinatorOrSelf, DetailView):
             if self.request.user.editor == editor and not editor.contributions:
                 messages.add_message(self.request, messages.WARNING,
                     # Translators: This message is shown on user's own profile page, encouraging them to make sure their information is up to date, so that account coordinators can use the information to judge applications.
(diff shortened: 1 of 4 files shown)


TWLight/users/views.py
```diff
@@ -92,8 +92,11 @@ def get_context_data(self, **kwargs):
             if self.request.user.editor == editor and not editor.contributions:
                 messages.add_message(self.request, messages.WARNING,
                     # Translators: This message is shown on user's own profile page, encouraging them to make sure their information is up to date, so that account coordinators can use the information to judge applications.
-                    _('Please update your contributions to Wikipedia (below) to '
-                      'help coordinators evaluate your applications.'))
+                    _('Please <a href="{}">update your contributions</a> '
```
jsnshrmn: Yeah for the contribution link, go ahead and assign and define a variable for substitution. Like:
https://github.com/WikipediaLibrary/TWLight/blob/3aec1fd1fc328887075f3ef322d2d15a447f9eb2/TWLight/views.py#L78

Samwalton9:
Ah, thanks! Fixed.

PR #92, opened 2018-05-21
Stop tag symbol showing on partner pages if they have no tags (T194638)

Samwalton9:


jsnshrmn:
Yeah, I believe this was a regression that I introduced. Thanks for fixing!

PR #93, opened 2018-05-21
Dashboard revamp

jsnshrmn:
pr#84 got hosed somehow, so I killed it and am recreating.

### DIFF
--- a/TWLight/graphs/views.py
+++ b/TWLight/graphs/views.py
@@ -13,7 +13,7 @@ from django.utils.translation import ugettext as _
 
 from TWLight.applications.models import Application
 from TWLight.resources.models import Partner
-from TWLight.users.models import Editor
+from TWLight.users.models import UserProfile, Editor
 
 from .helpers import (get_application_status_data,
                       get_data_count_by_month,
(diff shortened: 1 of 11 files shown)


jsnshrmn:
@Samwalton9 This looks about right to me. Et tu?

Samwalton9:
Yep, looks good! Thanks for fixing the PR :D

TWLight/graphs/views.py
```diff
@@ -345,3 +378,20 @@ def _write_data(self, response):
...
+        csv_queryset = UserProfile.objects.all()
+
+        data = get_user_language_data(
+            csv_queryset,
+            data_format=PYTHON
+        )
+
+        writer = csv.DictWriter(response, fieldnames=['label', 'data'])
+
+        writer.writerow({'label': _('Language'),
```
jsnshrmn: It might not be a bad idea to do a
```
.encode('utf-8')
```
on the language to prevent the server errors we have seen on some of the downloads occasionally. That's what I ended up doing with the paths.

Samwalton9: `get_user_language_data` already uses force_unicode on language (graphs/helpers.py L134).

jsnshrmn: duh

PR #94, opened 2018-05-22
Add ability for users to restrict processing of their data (T191880)

Samwalton9:
This PR also includes a couple of fixes for testing, including EditorFactory now assigning a random username so that the coordinator check doesn't think everyone is a coordinator when testing.


TWLight/applications/views.py
```diff
@@ -809,13 +811,23 @@ class ListReadyApplicationsView(CoordinatorsOnly, ListView):
     template_name = 'applications/send.html'
 
     def get_queryset(self):
+        # The annotate and filter functions filter out any partners who only
```
jsnshrmn: This warrants a bit more explanation/commentary. Please make this slightly more human parse-able even if it makes the code somewhat less concise.

TWLight/users/factories.py
```diff
@@ -29,7 +29,10 @@ class Meta:
         model = User
         django_get_or_create = ('username',)
 
-    username = 'alice'
+    # Multiple users with the same username can cause issues with group
```
jsnshrmn: Please take advantage of factory boy's use of faker when you need random test data, like:
https://github.com/WikipediaLibrary/TWLight/blob/ae1ae76eea600d9ac747fe560de907d0e4f3ee2c/TWLight/resources/factories.py#L14

Samwalton9:
Updated :)

TWLight/applications/views.py
```diff
@@ -809,13 +811,33 @@ class ListReadyApplicationsView(CoordinatorsOnly, ListView):
...
+        # Base queryset of partners with approved applications
+        base_queryset = Partner.objects.filter(
+                            applications__status=Application.APPROVED
+                            )
+
+        # For each partner in the queryset, count how many applications are
+        # from users with data restriction turned on
+        annotated_queryset = base_queryset.annotate(
+                                not_restricted_count = Count(
+                                    Case(When(~Q(applications__editor__user__groups__name='restricted'), then=1),
```
jsnshrmn: This line in particular is pretty dense. I had to look up Django's (Case expressions)[https://docs.djangoproject.com/en/1.11/ref/models/conditional-expressions/#case], (When objects)[https://docs.djangoproject.com/en/1.11/ref/models/conditional-expressions/#when], and (Q objects)[https://docs.djangoproject.com/en/1.11/topics/db/queries/#complex-lookups-with-q-objects] to understand exactly what it was doing. Considering that this is the only place in the application that those are used, I think it warrants more explicit commentary. You could break this single line up and add those comments. Also, since there's only one When in your Case, I think Case might be extraneous. That's an untested accusation, you might need the None it returns as an else condition.

jsnshrmn:
@Samwalton9, have I made a messy review for you to go through. Basically, your changes looked good, but I also wanted something specific that I hadn't asked for. First review approved, inline comments added (which attached to the first review), then a second review added (now with no inline comments). #winning!

Samwalton9:
Updated that whole block of code. Had a realisation that there was a vastly better way to generate that list.

jsnshrmn:
Way better!

PR #97, opened 2018-05-31
Allow users to delete their data

Samwalton9:
This is a big one and one that I'm more concerned about getting it right than restrict processing - take your time reviewing it!

I opted for keeping the application objects and just filtering out apps from deleted users rather than storing the data elsewhere because it seemed easier. We were going to want to keep almost all the non-personal metadata about an application, so it seemed reasonable to just keep the object itself. If there are issues with this method (which can include ease of use going forward) then that's totally understandable, don't worry about telling me to drop this method and actually figure out the 'track stats' method.


Samwalton9:
Been poking around with this today, some things I've noticed.

User comments don't get deleted.

When coordinators delete their accounts some unexpected things happen with applications they reviewed. They stop being listed on the Review views and navigating to them directly raises a 404. This should only be the case if the _submitting_ editor deletes their account.

Samwalton9:
I think I've addressed everything we talked about, except for unrelated email tests, which I filed as a Phab task for us to tackle separately :)

jsnshrmn:
We've got conflicting migrations from https://github.com/WikipediaLibrary/TWLight/pull/105

Samwalton9:
We've got the perennial also-includes-jasons-changes problem, but I should have fixed the migrations conflict.

jsnshrmn:
Yeah, the comparison that's not connect to the pr looks right
https://github.com/WikipediaLibrary/TWLight/compare/master...Samwalton9:delete-data

and I can see that your fork is up to date with master too
https://github.com/Samwalton9/TWLight/compare/delete-data...WikipediaLibrary:master

In previous cases I've closed the pr and created a new one. But I'm going to merge this one in and see what happens.

PR #101, opened 2018-06-04
Allow users to hide their applications from Latest Activity (T193776)

Samwalton9:
Couldn't see a good reason to leave the email text on applications because this appeared to be just about the only reason to apply via email.


jsnshrmn:
Conflicting migrations.

jsnshrmn:
FYI, I did a conflict resolution on ff to exclude hidden OR deleted apps from the timeline.

Samwalton9:
Aha - rebasing is the solution to jasons-work commit sneaking in :D

jsnshrmn:
What's strange is that it really shoulnd't be needed. The diffs between the two forks are correct outside of the PR. I'd call it a github PR implementation quirk.

PR #107, opened 2018-06-14
Add link to privacy policy in About You section of application form (T192746)

Samwalton9:


jsnshrmn:
I'm not sure that injecting html into the forms module is the best way to do this. It would be better to add this as a (translatable) message in a template.

Samwalton9:
I struggled to find a sensible solution to this because of the way that crispy forms is an all encapsulated thing in the template.

Do you know how I'd go about this without injecting HTML into the form?

TWLight/applications/forms.py
```diff
@@ -200,6 +201,10 @@ def _add_user_data_subform(self, user_data):
                 user_data_layout.append(datum)
 
             self.helper.layout.append(user_data_layout)
+            self.helper.layout.append(HTML('<p><small><i>Your personal data '
```
jsnshrmn: here's a reasonable way to get this bit of the interface captured for translation without reworking the way we're building the forms:
```
            # Translators: This this note appears in a section of a form where we ask users to enter info (like country of residence) when applying for resource access.
            self.helper.layout.append(HTML(_('<p><small><i>Your personal data '
                'will be processed according to our <a href="{terms_url}">'
                'privacy policy</a>.</i></small></p>'.format(
                    terms_url=reverse('terms')))))
```

jsnshrmn:
Can you rebase? the diff looks hinky.

Samwalton9:
Looks better?

PR #110, opened 2018-06-29
Update the homepage partner count automatically

Samwalton9:
Just a small change but will save us having to make pull requests every time this number updates.


jsnshrmn:
This should probably exclude unavailable partners from the count?

Samwalton9:
In the resources model unavailable partners are [hidden by default](https://github.com/WikipediaLibrary/TWLight/blob/master/TWLight/resources/models.py#L129).

jsnshrmn:
derp

PR #116, opened 2018-07-17
Allow designated coordinators to view a table of approved/sent users for their partner (T170795)

Samwalton9:
I know we'd talked about using a reporting tool, but not sure how far that progressed, so put something simple together!

### DIFF
--- a/TWLight/resources/templates/resources/partner_users.html
+++ b/TWLight/resources/templates/resources/partner_users.html
@@ -1,0 +1,86 @@
+{% extends "base.html" %}
+{% load i18n %}
+
+{% block content %}
+  {% comment %} Translators: This is the heading of a page listing editors who have applications for this partner ({{ object }}). {% endcomment %}
+  <h1>
+  {% blocktrans with partner=object.company_name %}
+  {{ object }} approved users
+  {% endblocktrans %}
+  </h1>
+
+  <div class="alert alert-info">
+    {% comment %} Translators: This message is shown on pages where coordinators can see personal information about applicants. {% endcomment %}
+    {% trans 'Coordinators: This page may contain personal information such as real names and email addresses. Please remember that this information is confidential.' %}
+  </div>
+
+  {% if approved_applications|length > 0 %}
+    {% comment %} Translators: On the page listing approved users for a partner, this is the title of the section listing applications with the 'approved' status {% endcomment %}
+    <h2>{% trans "Approved applications" %}</h2>
+    <table class="table table-sm">
+      <thead>
+        <tr>
+          <th>User</th>
+          <th>Email</th>
+          <th>Application date</th>
+          <th>Application approved</th>
+          <th>Renewal?</th>
+          {% if partner_streams %}
+            <th>Stream</th>
+          {% endif %}
+        </tr>
+      </thead>
+      <tbody>
+      {% for application in approved_applications %}
+        <tr>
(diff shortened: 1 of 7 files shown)


TWLight/resources/templates/resources/partner_users.html
```diff
@@ -0,0 +1,86 @@
...
+
+  <div class="alert alert-info">
+    {% comment %} Translators: This message is shown on pages where coordinators can see personal information about applicants. {% endcomment %}
+    {% trans 'Coordinators: This page may contain personal information such as real names and email addresses. Please remember that this information is confidential.' %}
+  </div>
+
+  {% if approved_applications|length > 0 %}
+    {% comment %} Translators: On the page listing approved users for a partner, this is the title of the section listing applications with the 'approved' status {% endcomment %}
+    <h2>{% trans "Approved applications" %}</h2>
+    <table class="table table-sm">
```
jsnshrmn: the text in this table should be marked for translation. Have you looked at this on a phone? Make use of a responsive grid instead of a table wherever possible.

TWLight/resources/templates/resources/partner_users.html
```diff
@@ -0,0 +1,86 @@
...
+        </tr>      
+      {% endfor %}
+      </tbody>
+    </table>
+  {% endif %}
+
+  {% if sent_applications|length > 0 %}
+    {% comment %} Translators: On the page listing approved users for a partner, this is the title of the section listing applications with the 'sent' status {% endcomment %}
+    <h2>{% trans "Sent applications" %}</h2>
+    <table class="table table-sm">
```
jsnshrmn: the text in this table should be marked for translation. Have you looked at this on a phone? Make use of a responsive grid instead of a table wherever possible.

TWLight/view_mixins.py
```diff
@@ -118,6 +119,46 @@ def dispatch(self, request, *args, **kwargs):
 
 
 
+class PartnerCoordinatorOnly(object):
```
jsnshrmn: Seems like we could have this and the coordinator or self class share more code. Perhaps a shared function for partner coordinators or maybe a class with a few subclasses?

Samwalton9:
Thanks!

- Made those strings translatable (can't believe I keep forgetting to do that)
- Added a `table-responsive` div to the table, per https://getbootstrap.com/docs/4.0/content/tables/#responsive-tables. Tested with Chrome's mobile view and it seems to be working much better, table is specifically horizontally scrollable, and the header doesn't get messed up at small widths anymore.
- Your point about consolidating the mixins made me realise the partner coordinator check could just slip right in to the existing `CoordinatorOrSelf`. Tested and seems to be working fine.

jsnshrmn:
So does that mean that editors can now download their own data on partner pages? That seems like it wouldn't be a problem, though it might be a bit of clutter for them. Also is partner_coordinator_only in twlight_perms.py being used anywhere anymore?

Samwalton9:
> So does that mean that editors can now download their own data on partner pages?

No - CoordinatorOrSelf is a bit misleading there. It checks if the view's get_object() output is an Editor, Application, or (now) Partner and corresponds to the user accordingly. Since the new view returns a Partner, the editor self check doesn't run. I've added this to the test though.

And nope, removed that function.

jsnshrmn:
Looks great!

PR #123, opened 2018-07-28
Show coordinators the total number of accounts available (T189608)

uyscuti-wiki:
The number of accounts available for a partner/stream(s) now available under the _Metrics_ section. On the _send page_, coordinators will get a warning if the accounts ready to send exceeds the number of available accounts.


uyscuti-wiki:
Line 51 in `send_partner.html` is a bit convoluted with the translation tags. Would love to know if there are more straightforward ways to go about this.

TWLight/resources/templates/resources/partner_detail.html
```diff
@@ -441,7 +441,29 @@ <h2>{% trans "Metrics" %}</h2>
...
+        {% comment %} Translators: This is the section header for accounts availability numbers for the available stream(s). (e.g. https://wikipedialibrary.wmflabs.org/partners/52/). {% endcomment %}
+        {% trans "Accounts availability" %}
+      </h3>
+      {% for stream_name, total_accounts in total_accounts_available_current.items %}
+        <p>{{ stream_name }}: {{ total_accounts }}</p>
+      {% endfor %}
+    {% else %}
+      <h3>
+        {% comment %} Translators: This is the label and the number which shows the number of accounts available for the particular partner. (e.g. https://wikipedialibrary.wmflabs.org/partners/8/). Don't translate {{ accounts_available }}. {% endcomment %}
+        {% blocktrans with object.accounts_available|add:negation_total_apps_approved_or_sent as accounts_available %}
```
Samwalton9: This should be `blocktrans trimmed with...` so as not to add a whole bunch of whitespace to the .po file.

TWLight/resources/views.py
```diff
@@ -60,6 +60,25 @@ def get_context_data(self, **kwargs):
...
+        if partner_streams.count() > 0:
+            context['total_accounts_available_current'] = {}
+            for stream in partner_streams:
+                context['total_apps_approved_or_sent_stream'] = User.objects.filter(
+                                      editor__applications__partner=partner,
+                                      editor__applications__status__in=(Application.APPROVED, Application.SENT),
+                                      editor__applications__specific_stream=stream).count()
+                
+                total_accounts_available = stream.accounts_available
+                context['total_accounts_available'] = total_accounts_available
```
Samwalton9: Seems like this is only used below here in the views.py file, and not in the template, so doesn't need to be part of the context dictionary; can just be a normal variable.

TWLight/resources/views.py
```diff
@@ -60,6 +60,25 @@ def get_context_data(self, **kwargs):
             partner=partner, status=Application.SENT).count()
 
         context['total_apps_approved_or_sent'] = context['total_apps_approved'] + context['total_apps_sent']
+        
+        # This if else block supports the template with the number of accounts available 
+        partner_streams = Stream.objects.filter(partner=partner)
+        if partner_streams.count() > 0:
+            context['total_accounts_available_current'] = {}
+            for stream in partner_streams:
+                context['total_apps_approved_or_sent_stream'] = User.objects.filter(
```
Samwalton9: This is a little confusing. It's a context per stream, but because it's in a for loop all that's passed to the template is the count for the last stream. It should broadly work as intended, but is a little confusing from a code point of view, and actually has a chance of not working as intended if there are no approved or sent applications for that last stream. 

Instead, consider making this a normal variable and turning partner_streams (Line 148) into a context variable to use as a check in the template.

TWLight/resources/templates/resources/partner_detail.html
Samwalton9: You should also do the addition in views.py rather than the template for clarity.

uyscuti-wiki:
let me know if something is amiss

TWLight/applications/views.py
```diff
@@ -867,11 +868,44 @@ def get_context_data(self, **kwargs):
...
                 )
         app_outputs = {}
-
+        stream_outputs = []
+        unavailable_streams = []
+        
         for app in apps:
             app_outputs[app] = get_output_for_application(app)
-
+            stream_outputs.append(app.specific_stream)
```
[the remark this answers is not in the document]
uyscuti-wiki: Ignored

TWLight/resources/templates/resources/partner_detail.html
```diff
@@ -441,7 +441,29 @@ <h2>{% trans "Metrics" %}</h2>
       </div>
     </div>
   </div>
-
+  
+  {% comment %}
+    Display # of accounts available
```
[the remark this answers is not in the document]
uyscuti-wiki: Interesting information, yes. But this pull is complicated with 29 commits as it stands, I'd suggest a new Phab task for this.

TWLight/applications/views.py
```diff
@@ -867,11 +868,53 @@ def get_context_data(self, **kwargs):
...
+        # Additionally, supports send_partner template with total approved/sent applications.
+        partner = self.get_object()
+
+        total_apps_approved = Application.objects.filter(
+            partner=partner, status=Application.APPROVED).count()
+
+        total_apps_sent = Application.objects.filter(
+            partner=partner, status=Application.SENT).count()
+
+        total_apps_approved_or_sent = total_apps_approved + total_apps_sent
```
Samwalton9: Apologies that this has only just occurred to me, but including approved apps here duplicates what's listed on the current send screen (applications listed at /send are 'approved'). So if you have 10 accounts available, 6 sent apps, and 3 approved apps, the platform will think you're out of accounts because 9 (approved or sent) + 3 (listed) > 10.

Really you don't need to count what's listed on send, you can just total 'approved or sent' and see if that's greater than the total available.

uyscuti-wiki: can't believe I missed this. Should be fixed now

PR #126, opened 2018-08-01
Added logging of personal data access for coordinators

Samwalton9:
Not sure I put everything in the right place here, so would appreciate a double check on that. Functionality works as intended.


Samwalton9:
I'll come back to this, apparently I forgot to run tests.

PR #127, opened 2018-08-01
Added script for running example data management commands

Samwalton9:
From https://github.com/WikipediaLibrary/twlight_puppet/pull/3.


Samwalton9:
This seems uncontroversial, merging in so I can get data deletion requests sent out.

PR #130, opened 2018-08-08
Add partner field for video tutorial links (T201152)

uyscuti-wiki:
Added a text field to display additional resources, if any, on `partner_detail.html`. I'm not quite sure why my migration field holds an `AlterField` to _status_, despite I've made no changes to it


TWLight/resources/templates/resources/partner_detail.html
```diff
@@ -196,6 +196,32 @@ <h1>{{ object }}</h1>
...
       </div>      
     </li>
+    {% if object.additional_resources %}    
+      <li class="timeline-inverted">
+        <div class="timeline-badge"><i class="fa fa-file"></i>
+        </div>
+        <div class="timeline-panel">
+          <div class="timeline-body">
+            <p>
+            {% comment %}
```
Samwalton9: This isn't really anything for you to change, but we should look into this and see if this comment is still accurate, or if the issue has been fixed, given that it's over a year old.

TWLight/resources/migrations/0049_auto_20180808_1625.py
```diff
@@ -0,0 +1,45 @@
...
+
+    operations = [
+        migrations.AddField(
+            model_name='partner',
+            name='additional_resources',
+            field=models.TextField(blank=True, help_text='Optional help resources for this partner.', null=True),
+        ),
+        migrations.AddField(
+            model_name='partner',
+            name='additional_resources_ar',
```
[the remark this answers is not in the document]
uyscuti-wiki: https://github.com/WikipediaLibrary/TWLight/blob/3e4b8a87c7ffbe5bafec2b7e0f5b8f059cd6deb2/TWLight/resources/migrations/0048_auto_20180713_1831.py :)

TWLight/resources/templates/resources/partner_detail.html
uyscuti-wiki: I did some digging and looks like the issue is [fixed](https://code.djangoproject.com/ticket/27882). Although, I did try using None with no luck. Update Django maybe?

uyscuti-wiki: Should be available in [version 2.0.x] (https://github.com/django/django/commit/7a7b331cd5975477597dac4dec7ee0ddb67f59e0), I think

Samwalton9: Ah, we're not going to be moving to 2.0 for a while yet, thanks for checking!

Samwalton9:
If you could fix the conflicts and migrations, then add to the example data script we should be good to go here too :)

uyscuti-wiki:
Done

Samwalton9:
Sorry, I had this comment half written and couldn't make my mind up enough to finish it! Two thoughts:

First, could we make this a translatable field? That way we can link to video guides in multiple languages depending on user language, and it will simply fall back to English if there's no need.

Second, let's make the row header 'Tutorial videos' (translatable), and simply have the videos be written as urls, comma separated. Then parse the list and number them as 'Video 1', 'Video 2', etc.

So if the field content was `https://youtube.com/abcd, https://youtube.com/efgh`, the view turns this into a list, then the template would have a for loop that for each url in the list writes the url HTML, labelling it 'Video {loop.counter}'.

Does that seem sensible?

We lose the per-video description ability, but I think supporting all that additional translation workload probably isn't worthwhile at this stage. We can improve it later.

uyscuti-wiki:
Should be fixed to show the URLs with the label 'Video #'

TWLight/resources/templates/resources/partner_detail.html
```diff
@@ -216,6 +216,26 @@ <h1>{{ object }}</h1>
...
+        <div class="timeline-panel">
+          <div class="timeline-body">
+            <h4 class="timeline-title">
+              {% comment %} Translators: If a partner has video tutorials, this text shows as the title for that information field.{% endcomment %}
+              {% trans "Video tutorials" %}
+            </h4>
+            <ul>
+              {% comment %} Translators: If a partner has video tutorials, this text links to the video(s).{% endcomment %}
+              {% for each_tutorial in video_tutorials %}
+                <li><a href={{ each_tutorial | safe }}>{% trans 'Video' %} {{ forloop.counter }}</a></li>
```
Samwalton9: I _think_ `{{ forloop.counter }}` should be inside the trans tag, to ensure that RTL languages still format the sentence correctly, but I'd appreciate if you could double check in Translatewiki's guidance for developers.

uyscuti-wiki:
Replaced trans with blocktrans and with

TWLight/resources/models.py
```diff
@@ -321,9 +325,9 @@ def save(self, *args, **kwargs):
           send_instructions_cache_key = make_template_fragment_key(
               'partner_send_instructions', [code, self.pk]
           )
-          cache.delete(short_description_cache_key)
```
jsnshrmn: Why have we changed up the order of cache delete operations here?

TWLight/resources/models.py
```diff
@@ -206,6 +206,10 @@ class Meta:
         "description such as collections, instructions, notes, special "
         "requirements, alternate access options, unique features, citations notes."))
         
+    additional_resources = models.TextField('video tutorials', blank=True,
```
jsnshrmn: The naming here bothers me.  We're calling one thing a resource, a tutorial, and a url in various places. Pick one and stick with it. If this is only ever going to be for tutorial video urls, then call the field tutorial_video_urls or something use the built in url validator https://docs.djangoproject.com/en/1.11/ref/validators/#urlvalidator

If you're thinking about being flexible for including resources other than video links, then break this out into its own model, and give it separate fields for url, resource type, and purpose, eg [url, video, tutorial], [url, text, faq], etc.

Samwalton9: Regarding the URL validator - the idea was that this field could accept a comma separated list of videos. Would the validator work in this case?

jsnshrmn: oh, yeah, that's icky. You'd then be splitting that string down the road, right? That's basically disregarding Django as an ORM.  If you want multiple things, you should have multiple objects. This should be its own model.

Samwalton9: That would indeed make more sense.

uyscuti-wiki:
Video is its own model now.

jsnshrmn:
This is looking really good. I'm going to pull it down and run it.

PR #133, opened 2018-08-12
Give partners 'short' and 'long' descriptions (T169618)

uyscuti-wiki:
Two types of descriptions, with `partner_tile`  displaying the short one and `partner_detail` displaying both. This is not perfect yet - submitting to see what could be added/changed.


uyscuti-wiki:
I was wondering, do we need filters for the long description now that we are restricting it to only the details page? If yes, we can show the expanded view by default and give users the option to hide, except that which are important.

TWLight/resources/models.py
```diff
@@ -190,10 +190,16 @@ class Meta:
         # Translators: In the administrator interface, this text is help text for a field where staff can link to a partner's Terms of Use.
         help_text=_("Link to terms of use. Required if users must agree to "
             "terms of use to get access; optional otherwise."))
-
-    description = models.TextField(blank=True, null=True,
-        # Translators: In the administrator interface, this text is help text for a field where staff can provide a description of a partner's available resources.
-        help_text=_("Optional description of this partner's resources."))
+            
+    short_description = models.TextField(max_length=1000, blank=True,
```
Samwalton9: You should create a custom migration file to rename description to short_description, and then add long_description to save us having to copy all the existing descriptions over to short_description.

See https://gist.github.com/dhbradshaw/e2bdeb502b0d0d2acced

TWLight/resources/templates/resources/partner_detail.html
```diff
@@ -175,14 +175,12 @@ <h1>{{ object }}</h1>
...
-                {{ object.description | twlight_wikicode2html | safe }}
-              {% endcache %}
-            {% endfilter %}
+            {% cache 31536000 partner_short_description LANGUAGE_CODE object.pk %}
+              {{ object.short_description | twlight_wikicode2html | safe }}
+            {% endcache %}
             </p>
           {% else %}
-            {% comment %} Translators: If a partner has no description written, this message is shown in the Description field on the partner page. {% endcomment %}
+            {% comment %} Translators: If a partner has no short_description written, this message is shown in the Description field on the partner page. {% endcomment %}
```
Samwalton9: short_description is an internal variable name and doesn't need to be shown to translators. 'description' is fine here.

TWLight/resources/templates/resources/partner_detail.html
```diff
@@ -175,14 +175,12 @@ <h1>{{ object }}</h1>
               We filter out details and summary tags so that we always see
               the full content on the details page.
             {% endcomment %}
-            {% filter twlight_removetags:"details summary" %}
```
Samwalton9: Is this filter no longer needed? If so, the comment above should be removed if it's no longer relevant.

uyscuti-wiki: Removed comment. I reckon we won't be using the details and summary tag in the short description (eg., EBSCO)

uyscuti-wiki:
@Samwalton9 was unable to use example_data. If the code looks ok, can you deploy this in staging?

uyscuti-wiki:
25 partners to have fake long descriptions

Samwalton9:
This is look great Aaron! I've just pulled it to do a final check and noticed the long description is going into a box of its own in the sidebar of partner pages. I'm hesitant to add too many more of those - do you think you could instead put the long description inside the main description box, just under the tags?

PR #136, opened 2018-08-21
WIP Scale graph ticks appropriately to avoid overlap. Also fixed graph hiding bug.

Samwalton9:


Samwalton9:
Literally moments after filing this PR I realised that this is partially default flot behaviour. We were overwriting tickSize, but if you leave that argument out it scales sensibly, including with window size. The only thing that needs fixing up is one day of data showing 00:00:00 instead of, I guess, not showing. Curious what it looks like with a few days worth of data. Ignore this PR for now.

jsnshrmn:
just following up. Can this be closed out?

PR #137, opened 2018-08-30
Provide a link or form for suggesting new partners (T201156)

uyscuti-wiki:
A new form with name, description and URL, followed by listed suggestions with delete option for coordinators.


uyscuti-wiki:
Will generate example data once this is good to go.

uyscuti-wiki:
Now with upvote option for suggestions. Although I'm not sure the toggle of translated 'Upvoted'/'Upvote' would pan out; would love some inputs.

brodieaustin:
One suggestion to consider--the button expands right and night overflow the box if the number gets to be double (or even triple digit). You might consider:

```
<div class="col-md-2 pull-right" style="text-align: right;">
     <a class="btn upvote btn-default" data-href="/suggest/1/upvote/" href="">
             <span class="utext">Upvote </span><span class="count badge">2</span>
     </a>
</div>
```
This will give the button room to grow but push the button right.

uyscuti-wiki:
can be tested in [staging](http://twlight-staging.wmflabs.org/suggest/)

uyscuti-wiki:
prominent button on 'partner filter' page to reach suggest page

Samwalton9:
Looks good to me though I'd appreciate a final confirmation from @jsnshrmn and we should wait for the translation fix before merging.

uyscuti-wiki:
added example data

TWLight/resources/models.py
```diff
@@ -432,4 +432,52 @@ class Meta:
 
     def __unicode__(self):
         # As with Stream, do not return the partner name here.
-        return self.full_name
\ No newline at end of file
+        return self.full_name
+
+
+
+class Suggest(models.Model):
```
jsnshrmn: This is a nitpick, but please call this a Suggestion (noun), rather than Suggest (verb). All of our models represent things not actions.

TWLight/resources/models.py
```diff
@@ -432,4 +432,52 @@ class Meta:
...
+        
+    suggested_company_name = models.CharField(max_length=40,
+        # Translators: In the administrator interface, this text is help text for a field where staff can add partner suggestions.
+        help_text=_("Potential partner's name (e.g. McFarland)."))
+
+    description = models.TextField(blank=True, max_length=1000,
+        # Translators: In the administrator interface, this text is help text for a field where staff can provide a description of a potential partner.
+        help_text=_("Optional description of this potential partner."))
+
+    company_url = models.URLField(blank=True, null=True,
```
jsnshrmn: let's use the url validator here.
https://docs.djangoproject.com/en/1.11/ref/validators/#urlvalidator

TWLight/resources/models.py
```diff
@@ -432,4 +432,52 @@ class Meta:
...
+        # Translators: In the administrator interface, this text is help text for a field where staff can link to a potential partner's website.
+        help_text=_("Link to the potential partner's website."))
+    
+    author = models.ForeignKey(User, related_name='suggest_author', blank=True, null=True, on_delete=models.SET_NULL,
+        # Translators: In the administrator interface, this text is help text for a field where staff can link a user as the author to a suggestion.
+        help_text=_("User who authored this suggestion."))
+    
+    plus_ones = models.ManyToManyField(User, related_name='suggest_likes', blank=True,
```
jsnshrmn: I don't like the wording naming/help text/relationship here. Does this represent a user or a vote? Why are we switching between terminology for user votes (likes vs +1s)

TWLight/resources/models.py
```diff
@@ -432,4 +432,52 @@ class Meta:
...
+    
+    def get_upvote_url(self):
+        return reverse('upvote', kwargs={'pk': self.pk})
+    
+    
+    def save(self, *args, **kwargs):
+        """Invalidate this suggestions's pandoc-rendered html from cache"""
+        super(Suggest, self).save(*args, **kwargs)
+        
+        description_cache_key = make_template_fragment_key(
```
jsnshrmn: The caching stuff is only needed if you're using our twlight_wikicode2html mixin for the description.

TWLight/resources/templates/resources/suggest.html
```diff
@@ -0,0 +1,107 @@
...
+                <p>
+                  <strong>{% trans 'Description' %}: </strong>
+                  {% comment %}
+                    Cache lifetime should be None, which would be forever since we're
+                    invalidating on update. However, as 2017-04-06 a bug in django
+                    doesn't allow for None to be passed from here into cache.  The
+                    fix is in master, but not release.
+                    https://code.djangoproject.com/ticket/27882
+                  {% endcomment %}
+                  {% cache 31536000 suggested_partner_description suggest.pk %}
```
jsnshrmn: This caching stuff is only need if you're using our twlight_wikicode2html mixin, eg ```suggestion.description | twlight_wikicode2html | safe```

TWLight/resources/templates/resources/suggest.html
```diff
@@ -0,0 +1,107 @@
...
+            </div>
+          </div>
+        {% endfor %}
+      </div>  
+    </div>
+  {% endif %}
+{% endblock content %}
+{% block javascript %}
+  <script>
+    $('.upvote').click(function(e){
```
jsnshrmn: again, is this a like, a plus one, an upvote? pick one.

TWLight/resources/templates/resources/suggest_confirm_delete.html
```diff
@@ -0,0 +1,15 @@
...
+{% block content %}
+  <form method="post">
+    {% csrf_token %}
+    <p>
+      {% comment %} Translators: This message is displayed on the page where coordinators can request the deletion of a suggestion. {% endcomment %}
+      {% blocktrans trimmed %}
+        Are you sure you want to delete <b>{{ object }}</b>?
+      {% endblocktrans %}
+    </p>
+    <input type="submit" value="Confirm" class="btn btn-danger"/>
```
jsnshrmn: Confirm should be marked for translation, no?

TWLight/resources/views.py
```diff
@@ -240,3 +245,81 @@ def get_context_data(self, **kwargs):
...
+        self.object = self.get_object()
+        self.object.delete()
+        messages.add_message(self.request, messages.SUCCESS,
+        # Translators: Shown to coordinators when they successfully delete a partner suggestion
+        _('Suggestion has been deleted.'))
+        return HttpResponseRedirect(self.success_url)
+
+
+
+class SuggestUpvoteView(EditorsOnly, RedirectView):
```
jsnshrmn: upvote, plus one, like?

TWLight/resources/models.py
uyscuti-wiki: sorry, but I'm a bit lost. I thought the point of using a `URLField` was to not have to use `urlvalidator`

jsnshrmn: uh, yep, you can disregard that. My brain made that a charfield for no good reason.

uyscuti-wiki:
Although I made sure to not miss anything from your comments, it's possible I might've missed something. Apologies in advance if that's the case.

TWLight/users/migrations/0031_auto_20181010_1744.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
uyscuti-wiki: I made no changes to the UserProfile model, and am not sure why I have an AlterField migration on the same

jsnshrmn: You picked up a migration caused by languages being added by wiki translate. Go ahead and delete this file from your branch, as I have discovered that I need to go scoop those migrations out of the production branch and push them back to master so that we don't end up in migration hell.

jsnshrmn:
The changes look good, and are consistent with what I was asking for. I fast forwarded this and resolved the conflicts so I could run it and do another review pass.

TWLight/resources/views.py
```diff
@@ -240,3 +245,81 @@ def get_context_data(self, **kwargs):
...
+            suggestion.upvoted_users.add(self.request.user)
+            messages.add_message(self.request, messages.SUCCESS,
+            # Translators: Shown to users when they successfully add a new partner suggestion.
+            _('Your suggestion has been added.'))
+            return HttpResponseRedirect(reverse('suggest'))
+        except AssertionError:
+            messages.add_message (request, messages.WARNING,
+                # Translators: This message is shown to users who attempt to post data to suggestion form without logging in
+                _('You must be logged in to do that.'))
+            raise PermissionDenied
```
jsnshrmn: I know we ended up with this after discussing how the display logic should work, but now that I see it in action, I think we might be better served by either:
a) wrapping the whole page in the login redirect machinery like everything else, or
b) wrapping just the form post like that.

jsnshrmn: @uyscuti-wiki I'm throwing a little bit of a wrench into some code that works exactly as we all discussed. If the thought of working on this any more makes you want to throw your hands up, let me know and I can bang out this piece.

uyscuti-wiki: I'm unsure as to how to proceed without some additional help. Don't think the first option would work since we are interested in allowing unauthenticated users to browse the existing suggestions. As for the second option, is it a different method I'd have to override? Or is it something else? 
If you think it's too much trouble to explain, I'm okay with "banging out this piece" (not sure what that is either) :)

jsnshrmn: Yeah you're right about option a not really being an option, which is why we landed where we are. For option b, you just need to make sure that the view that processes the form post has its own class-based url defined, then you can add the login_required decorator to that url. Here's an example from graphs:
https://github.com/WikipediaLibrary/TWLight/blob/c3834b5dd40c31e12db09672f240918cbfb0b5c9/TWLight/graphs/urls.py#L2
https://github.com/WikipediaLibrary/TWLight/blob/c3834b5dd40c31e12db09672f240918cbfb0b5c9/TWLight/graphs/urls.py#L43-L46

"banging it out" is an example of not-super helpful helpful idiom that I lapse into sometimes.
https://idioms.thefreedictionary.com/bang+out#Definition
Apologies.

uyscuti-wiki:
Heh, no problem. Thanks for the help. I'll give it a try.

uyscuti-wiki:
@jsnshrmn I've been working on this for a while now and I don't think I can pull this off without shaking-up my understanding of Django forms from the ground up. Can you take this over?

jsnshrmn:
@uyscuti-wiki sure thing. I'll do a walk-through with you when it's done so that we can still make it a learning opportunity.

jsnshrmn:
@uyscuti-wiki so, this isn't done yet, but I did get around to taking a pass at it today. What I landed on was squeezing the login_required decorator *inside* a post method decorator for the class based form view.

That made it so the login_redirect only happens on post, not get. But we got redirected without any of the data we entered into the form.

So I overrode get_initial for that view so I could add in some get parameter processing.

At that point we'd be in good shape, but for ouath. There's a little bit more work for me to do to make sure the form data survives the trip out to the identity provider and back.

I realize now that if you had made it this far, you almost certainly would have found the next step to be a stumbling block.

jsnshrmn:
@uyscuti-wiki I went ahead and did the housekeeping tasks.

PR #143, opened 2018-09-26
SUL Info link does not appear to work on any application (T202837)

uyscuti-wiki:
Removed all instance of SUL Info tool from the tool. Although I'm not sure about migrations since all the new languages are new fields, I wonder if there's an automated way of generating migrations.


uyscuti-wiki:
Also not sure if I had to make the changes to the translation files manually.

Samwalton9:
What needs to happen here - translation commits?

PR #144, opened 2018-09-27
Access codes functionality

Samwalton9:
This PR adds a staff dashboard where staff can upload .csv files containing access codes, which are stored in a new resources model AccessCode.

Partners are modified to have multiple distribution types, including placeholder proxy and bundle types (replacing the previous Yes/No bundle option). For partners distributing via access codes, coordinators have a new Send Partner interface, where codes are assigned to applications. Codes can be removed in the List Applications view for each partner.

In future we could add email text for partners and have the access codes send out automatically. For now, coordinators will still need to send the codes via email.


TWLight/resources/models.py
```diff
@@ -214,11 +230,12 @@ class Meta:
...
-          
-    bundle = models.NullBooleanField(
-        blank=True, null=True, default=False,
-        # Translators: In the administrator interface, this text is help text for a field where staff can specify whether users can access this as part of the Bundle.
-        help_text=_("Is this partner a part of the Bundle?"))
+
+    distribution_type = models.IntegerField(choices=DISTRIBUTION_CHOICES,
+        default=EMAIL,
+        # Translators: In the administrator interface, this text is help text for a field where staff can specify which method of account distribution this partner uses.
+        help_text=_("Which distribution method does this partner use? "
```
jsnshrmn: Seems like this should be a little more expository about how the distribution methods should work?
It might be noting that bundle is also proxied.

TWLight/settings/base.py
```diff
@@ -172,6 +173,11 @@ def get_languages_from_locale_subdirectories(dir):
 
 SITE_ID = 1
 
+# Overwrite messages.ERROR to use danger instead, to play nice with bootstrap
```
jsnshrmn: This seems reasonable, but why is it in this pr?

TWLight/templates/staff.html
```diff
@@ -0,0 +1,62 @@
+{% extends "base.html" %}
```
jsnshrmn: This seems like the kind of thing that we should be using the delivered admin tools for instead of writing our own interface. It is possible to give staff limited access to the admin site.  Are we needing something more granular than what can be done there? I think we've talked about parts of this before, but it would be worth revisiting.

TWLight/view_mixins.py
```diff
@@ -122,6 +126,29 @@ def dispatch(self, request, *args, **kwargs):
...
+class StaffOnly(object):
+    """
+    Restricts visibility to:
+    * Staff; or
+    * Superusers.
+    """
+
+
+    def test_func_staff_only(self, user):
+        return (user.is_superuser or user.is_staff)
```
jsnshrmn: I don't think you need to check for both? I think superuser passes the staff check. don't take word for that.

TWLight/views.py
```diff
@@ -165,3 +174,100 @@ def get_context_data(self, **kwargs):
...
+
+    def get(self, request, *args, **kwargs):
+        context = {}
+
+        context['accesscode_partners'] = Partner.objects.filter(
+            distribution_type=Partner.CODES)
+
+        return render(request, 'staff.html', context)
+
+    def post(self, request, *args, **kwargs):
```
jsnshrmn: This warrants a second pass for security/safety, since we're running files from the internet through the python interpreter.

TWLight/templates/staff.html
Samwalton9: Perhaps, though isn't this is a little complex for the admin dashboard, at least in it's current form? This parses data from a CSV to turn into objects, could we do that in the admin interface?

Additionally, I approached this from the basis that a staff dashboard for doing things like this was something we were discussing (i.e. this page could be expanded to include adding a new partner, assigning a coordinator, etc.). I appreciate that one of the other options we talked about was limited /admin access though.

TWLight/view_mixins.py
Samwalton9: Ah, interesting. Like the comment above, I approached this from our eventual aim of 'staff' and 'superuser' being somewhat more distinct than they are now.

Samwalton9:
@jsnshrmn Could you elaborate?

Also, did you have any thoughts on my question re: CSV uploads via /admin?

TWLight/templates/staff.html
jsnshrmn: Sorry for being slow to respond to this question. To answer:
* Yes, I think this functionality can be added in the admin dashboard.
* There is nothing wrong with this approach, but it implements a design decision that I wasn't aware we had reached. There be consequences.

TWLight/view_mixins.py
jsnshrmn: Where our code is using django.contrib.auth, a superuser is treated as a user with all permissions assigned:
https://docs.djangoproject.com/en/1.11/ref/contrib/auth/#django.contrib.auth.models.User.is_superuser

Staff just means they are able to see the admin site:
https://docs.djangoproject.com/en/1.11/ref/contrib/auth/#django.contrib.auth.models.User.is_staff

jsnshrmn:
@Samwalton9 Added more thoughts inline, and resolved things where your response cleared things up. Regarding authorization:
You are assigning access codes based on application status since that's what we currently have. There is another type of object that we'll need to have for proxy, which I'm calling an authorization. Those authorizations are what need to handle access management.

TWLight/templates/staff.html
Samwalton9: Note to self, something like: https://stackoverflow.com/questions/3974620/import-csv-data-into-database-in-django-admin

Samwalton9:
Quick thought - the authorization model should probably also have a stream ForeignKey, in case an editor is only approved for a particular collection rather than the partner as a whole.

jsnshrmn:
Definitely! I for things that need both, I always just start with partner and then add streams in when I'm happy with the way it works.

jsnshrmn:
@Samwalton9 I think this is ready for you to look at and for us to pick back up on the discussion. I now have completed applications creating authorizations, and access codes can be attached to those authorizations. My thought here is that we can now connect this up and have the access code be automatically sent to a user when available.

Samwalton9:
Note that something's gone wrong here, looks like I might have merged files badly. Will investigate soon.

Samwalton9:
Alright. I screwed up my Git-ing while trying to rebase and accidentally removed your intermediate commits @jsnshrmn. Luckily I had them stashed so they're now contained in my most recent commit. It looks like everything should be as intended overall?
If it doesn't look like anything's missing I'll resolve the merge conflicts with master.

Samwalton9:
Not sure why the example data scripts decided to show as entirely changed, but they did have changes, to generate example access codes for partners and link them to applications.

Samwalton9:
That looks to have improved things.

jsnshrmn:
This is much easier to parse now. Thanks! I think this looks good pending a resolution of the merge conflicts. If you can take care of that, then I'll pull this down again and hammer on it. Or vice versa.

jsnshrmn:
we should update ui stuff like the application list on the user profile page over to authorizations rather than apps

jsnshrmn:
I picked this back up and ran into an issue with our example data command for applications: Approved applications don't actually get approved by a user. Because of that, no authorizer can be set for the authorization object, which is problematic.

Samwalton9:
Hmm. Am I right in thinking that we determine _who_ approved an application by the object history rather than by recording it in the model?

jsnshrmn:
Yep, the issue is that the approver gets set based on the request user (e.g. a logged in coordinator) which isn't the workflow the management command uses. The most conceptually straightforward way to accomplish that would be to approve the applications with a django client (like we do in some tests) instead of directly updating the objects. It seems like there's probable a way we could hack it into the revision though.

jsnshrmn:
I went ahead and made the model a little more accepting of missing data, just to make working on this easier while we get this sorted.

jsnshrmn:
I got my issue sorted out with how the authorizations show up in the user view. Now the related info shows up in both places. The way I'm showing streams in the dropdown might not be ideal. Basically, it shows all streams, not just those related to a selected partner.

jsnshrmn:
Ah, I do need to drop migrations first though, as this pr predates our migration handling in the travis pipeline.

PR #189, opened 2018-12-03
Add matomo tracking support

jsnshrmn:
Should be good to go, but we can't merge it in quite yet. Requires a small update to [twlight_puppet](https://github.com/wikipedialibrary/twlight_puppet) so that we're setting values for matomo in the environment_vars.py files.


jsnshrmn:
I've removed the requirement for any puppet changes, though getting there caused me to improve the puppet module a little in the process.

PR #191, opened 2018-12-09
Add INVALID status option

arcayn:
This fixes [T175249](https://phabricator.wikimedia.org/T175249)

Added ability for application status to be "Invalid", removed these applications from metrics/data analysis.


arcayn:
Just to clarify does that mean to use `-danger` for the `INVALID` entry on the `LABELMAKER` in applications/models.py?

Thanks :)

Samwalton9:
I believe so - then the functions in [version_tags.py](https://github.com/WikipediaLibrary/TWLight/blob/master/TWLight/applications/templatetags/version_tags.py) should handle the display.

arcayn:
Maybe we could use a different colour (e.g. the secondary grey I've included so far) so that users can easily tell the invalid and rejected apps on the same page apart? Just wondering

Samwalton9:
Sure - grey could work :)

arcayn:
Does my solution to use '-secondary' not work then? I believe that boostrap class should have a grey colour and can be used anywhere '-primary' can be

Samwalton9:
Ah, I missed that you'd already added -secondary. I had pulled your changes down and checked it and there was simply no tag for Invalid applications. From a little digging I can see that's because we're using Bootstrap 3.3, and `secondary` wasn't added as a colour option until version 4. If you just go with the same colour as rejected for now we can update this to secondary in the future when we've updated to 4 :)

arcayn:
Ah I had no idea! Will change

Samwalton9:
This looks great! I wouldn't mind getting a 2nd look over by @jsnshrmn when he's back in January in case this has consequences I haven't considered.

If you're interested in continuing to contribute to the platform you can find further open tasks at https://phabricator.wikimedia.org/project/view/2765/ :)

PR #196, opened 2019-01-02
Improving developer documentation

Samwalton9:
* Wrote shell script documentation
* Notes on pushing to the repo
* Making strings translatable


Samwalton9:
Can't think of anything else that needs documenting in this file (though we still need better twlight_vagrant developer-focused docs), so I'd say this is good to go for now. No sense leaving it here until I think of something else :)

PR #199, opened 2019-01-03
T209836: Improve user check behaviour in shell scripts

jsnshrmn:
This should resolve the behavior described in:
https://phabricator.wikimedia.org/T209836

I never had the errant vagrant exiting behavior in debian 9 on docker, so I'll need a debian 8 vm tester.


Samwalton9:
Seems to be working correctly, ran a bunch of scripts, including activate, both as www and not. Everything seemed to work as expected - I didn't get kicked right out of the session when hitting the exits, and scripts ran correctly as www.

PR #200, opened 2019-01-04
Shell script cleanup

jsnshrmn:
This is the remainder of the shell script cleanup that should firm up the boundaries between the various scripts called for polling git for updates and applying them. I've moved the spot where backups are fired, so that backups are now only fired by the update script if there is a code change (so now we should be able to update much more frequently). I also decided to have failure emails contain the output of the failed commands (the same thing that gets written to the logs) to allow for faster troubleshooting.  update.log should now also contain test output.


Samwalton9:
Hard to test locally because of all the behind master, "${remote}" = "${base}", etc. checks. Looks sensible though - if you want to test before pushing to prod maybe easier to try Staging?

jsnshrmn:
need to add some logic to run backups at least once daily even if there is no code to deploy

jsnshrmn:
Okay, I added locking on backup and deploy tasks which will allow us to have a separate cron task for backups and deploys that won't overrun each other. As it is, deploys will still happen every hour, but now we have the ability to configure more deploys via the puppet module.

PR #206, opened 2019-01-09
Add missing URL tags to email html

Samwalton9:
A couple of our email HTML files didn't have app_url wrapped in <a> tags so they weren't rendering as links when received.


Samwalton9:
-0.002%...

PR #209, opened 2019-01-10
Streamline dependencies and update those used by mwoauth

jsnshrmn:
During the last round of dependabot PRs, I took a look at our dependency graph and realized that we weren't handling oauth dependencies correctly. This PR fixes that by moving to a mwoauth fork with updated dependencies and pulling them out of TWLight's requirements.txt.


jsnshrmn:
I tested this code manually by pulling it down and signing in and out. I changed my accepted language several times to verify our hacks still work. Looks good.

PR #211, opened 2019-01-16
Add a Contact Us page (T201159)

uyscuti-wiki:
A very _basic_ working model of the contact us page. Probably needs some more work before it can be merged, but this is all I've got on my own.


Samwalton9:
Nice! It's probably worth adding a test to ensure that the contact form fires an email correctly.

uyscuti-wiki:
Added a test case with my limited understanding of testing to trigger the email function linked to the contact us form.

Samwalton9:
Looking good! Just a minor thing - "comtact" rather than "contact" :D

Samwalton9:
>If the user isn't signed in, we then throw away their input for the email field after sign in

This doesn't seem to be the case for me - if I click Submit while logged out, I get logged in and returned to the page with my form data intact.

Samwalton9:
Oh, right - I see that, yep.

Honestly I like the first idea - we can display a 'you need to go set your email' message if they haven't set it.

jsnshrmn:
I think ideally we'd still display the email value (after sign in) as a locked/grey-out field (my recollection is that crispy makes this simple) with some help text stating that this comes from the profile.

uyscuti-wiki:
The email field no longer is directly editable by users, both logged in and not logged in.

Samwalton9:
@uyscuti-wiki Did you mean to make your last commit in this PR? Feels like its unrelated to the Contact page, unless I'm missing something.

uyscuti-wiki:
About the first point, I thought so too, until I came across (this)[https://docs.djangoproject.com/en/1.11/ref/forms/fields/#disabled] - Django won't let users tamper with the data. What am I missing here?

jsnshrmn:
@uyscuti-wiki I stand corrected on that point! You're not missing anything, this is Django doing the right thing!  Since we're not actually using that email address anywhere yet, I couldn't actually see the parsed value. I just modified it and saw that it succeeded.

jsnshrmn:
@uyscuti-wiki Do you want to work on any more steps for this or would you like to hand off?

uyscuti-wiki:
Not yet. I'll give it a whirl, see if I get anywhere. On a related note, can you point me to documentation/guides (if any) on djmail? Thanks for asking.

jsnshrmn:
django email docs:
https://docs.djangoproject.com/en/1.11/topics/email/

djmail docs:
https://bameda.github.io/djmail/

TWLight/emails/views.py
```diff
@@ -41,17 +42,26 @@ def form_valid(self, form):
...
-            ContactUs.new_email.send(
-                sender=self.__class__,
-                user_email=email,
-                editor_wp_username=self.request.user.editor.wp_username,
-                body=message
-            )
-            messages.add_message(self.request, messages.SUCCESS,
-            # Translators: Shown to users when they successfully submit a new message using the contact us form.
-            _('Your message has been sent. We\'ll get back to you soon!'))
+            # Additional validation on the email
```
jsnshrmn: Two thoughts:
1. This should use django's builtin email validator
    https://docs.djangoproject.com/en/1.11/ref/validators/#emailvalidator
2. It should probably happen on the profile page, since that's the place we're actually accepting this input.

uyscuti-wiki: I've [moved the validation over to users app and made use of the built-in validator](https://github.com/WikipediaLibrary/TWLight/pull/211/commits/24078aec926d0c1baa175270fca8259005e120e8). Couple of question though: 

1. the code validates the email field data on `email_change` form submit. Is that how you pictured it happening? If not, I have no clue as to when or how to validate email from the user's profile page. 
2. the email field of the said form and the email field in `models.py` both happen to be EmailFields and use Django's built-in validators. Assuming what I did was right, why do we need this additional validation on top of that?

jsnshrmn: Yeah, I didn't actually go back through and check the model. Consider my feedback to be declarative, not imperative "ensure we have email validation" rather than "implement email validation". That was my bad. Let's not double-implement this.

uyscuti-wiki: No problem! Additional validation is no more.

TWLight/emails/templates/emails/contact_us_email-subject.html
```diff
@@ -0,0 +1,5 @@
+{% load i18n %}
+{% comment %} Translators: This is the subject line of an email sent to the Library Card platform admins from the contact us form.{% endcomment %}
+{% blocktrans trimmed %}
+Library Card platform email from user "{{ editor_wp_username }}"
```
jsnshrmn: I manually tested usernames with utf-8 characters since a) those are allowed in wp usernames and b) they can be problematic in email. It looks like Djmail does the right thing on our behalf, or at least doesn't fall over and die. Let's change this subject line up a bit though:
```suggestion
Wikipedia Library Card Platform message from {{ editor_wp_username }}
```

uyscuti-wiki:
Thanks!

PR #216, opened 2019-01-24
Fix partner application graph ranges and labels (T171159)

uyscuti-wiki:
Related to #136. Restore default tickSize and barred graph from renderring should there be only a week's worth of data. @Samwalton9 is this what you were expecting?

### DIFF
--- a/TWLight/graphs/helpers.py
+++ b/TWLight/graphs/helpers.py
@@ -49,6 +49,20 @@ def get_median(values_list):
     return median
 
 
+def get_earliest_creation_date(queryset):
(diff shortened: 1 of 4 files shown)


Samwalton9:
Looks good to me - tested locally and seems to be working well. @jsnshrmn could you do a quick scan and merge if all looks good?

Samwalton9:
>couldn't we check the boolean imported field for application

No, some imported apps have dates, and some don't. Those that don't were given dates of 1970, but we'd like to keep the others in mind when deciding on the graph range.

PR #219, opened 2019-01-26
Don't show negative numbers for accounts available (T211022)

uyscuti-wiki:
Negative numbers are now visible to only coordinators and superusers.


uyscuti-wiki:
should be fixed

PR #220, opened 2019-01-26
Don't send users to /comments/post/ with a form error if they try to comment without an email set (T206614)

uyscuti-wiki:
Email no longer required to post comments on applications.


uyscuti-wiki:
added credits

PR #225, opened 2019-02-07
WIP - Set up database content translations on Meta

uyscuti-wiki:
Short descriptions can be pulled from meta for now.


TWLight/resources/management/commands/meta_translations.py
```diff
@@ -0,0 +1,62 @@
...
+                languages_on_revision_field = ast.literal_eval(each_partner.short_description_last_revision_ids)
+            for every_language in languages_on_file:
+                if every_language[0] not in languages_on_revision_field:
+                    languages_on_revision_field[every_language[0]] = None
+            each_partner.short_description_last_revision_ids = languages_on_revision_field
+
+            response = requests.get('https://meta.wikimedia.org/w/api.php?action=query&format=json&meta=messagegroupstats&mgsgroup=page-Library_Card_platform%2FTranslation%2FPartners%2FShort_description%2F{partner_pk}'.format(partner_pk=each_partner.id))
+            language_data = response.json()
+            if 'error' in language_data:
+                print 'No short_description for partner {}.'.format(each_partner)
```
Samwalton9: Could the page text have "error" in it but not have errored?

e.g. if the description text has the word "error" would it skip here?

Samwalton9:
Just an early thought that isn't in the spec we drew up for this - we should do some sanity checks on the incoming data, like making sure it's not excessively long. There are probably other worthwhile checks that I'm just not thinking of right now.

TWLight/resources/management/commands/meta_translations.py
uyscuti-wiki: The latter seems to be the case, the `in` keyword looks for keys.

jsnshrmn:
I really think we could skip the db and the management command all together for this content.

We could use a template fragment cache like we do for the current pandoc output, which is what the twlight_wikicode2html template tag is fetching. If you peek at the templates that call partner descriptions, you'll see how those are leveraging cache. Instead of calling something defined in a template tag, though, you can call data defined in the view, which is where you could stick your api call to fetch content.

If you wrap that api call in a try block, you can return the English version for any missing translations, which would mimic the current behavior for django-managed content.

For the current setup, cache invalidation is handled in the model, since we know the cache needs to be flushed when a new version of a partner description gets saved. We won't have an event like a model save to hook onto in this case, so we'll need to do something else to invalidate the cache.

One thought is to take your revision checking call and make it an asynchronous thing that gets fired when the view gets called. That way we can go ahead and render the current view with what we have in the cache and do checking/invalidation/flushing in the background.

One way to enable asnyc would be to set up celery, but then again we're using async email sending in djmail and we haven't configured celery for that. That's a place for a bit of research after doing the other restructuring.

uyscuti-wiki:
@jsnshrmn would you be able to take a quick peek and confirm if I'm headed in the right direction?

uyscuti-wiki:
Everything works except for the Celery integration i.e. the cache invalidation which is now a Celery task. And the partner browse page takes a solid 20 - 30 seconds to load (at least for the first time) due to the n number of serial API calls. Is there a way to make and process the calls parallelly?

uyscuti-wiki:
I'm also considering breaking out the API calls to services.py for legibility reasons.

uyscuti-wiki:
@jsnshrmn almost everything required of this task is done except for the Async integration piece and removing all instances of short, long and stream descriptions and ensuring nothing breaks.

uyscuti-wiki:
[Now available in staging](https://twlight-staging.wmflabs.org/partners/)

uyscuti-wiki:
Removed short, long and stream descriptions and [ran migrations](http://dpaste.com/1J9H2DB), only to find the fields still intact in the table. Do I have to resort to writing custom migrations?
screenshot 87

jsnshrmn:
I believe we're creating those fields by running the `sync_translation_fields` [modeltranslation management command](https://django-modeltranslation.readthedocs.io/en/latest/commands.html). You might try running it and see if it drops the translation fields now that the original field is gone.

uyscuti-wiki:
On executing `sync_translation_fields` I get `No new translatable fields detected` and the fields aren't dropped in the table.

jsnshrmn:
Then I think you'll need to write some migration code to drop translation fields for model fields that no longer exist.

jsnshrmn:
doing a quick rebase, as this fork was just barely too old to work in my dev environment.

jsnshrmn:
or rather, not. it looks like @uyscuti-wiki hasn't given write-back privileges on this pr.

jsnshrmn:
https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

jsnshrmn:
@uyscuti-wiki never mind me. I was on the old contact-us branch of your fork. Working from a completely irrelevant branch makes a difference!

jsnshrmn:
Now I've actually done a fast forward on this. I did a clone off to the side in case this causes any problems.

jsnshrmn:
Yeah, the state of the branch doesn't look quite right now. I'll force push the previous state and Terry again.

jsnshrmn:
Okay, we now have things rebased against master, which means I can do a vagrant up now and have it be reasonably worky.

jsnshrmn:
More thoughts: I'd probably stop using the template fragment cache altogether and directly manage the cache in the view code that's making the api calls since we have a good basis for setting our expiration interval at request time. Maybe we then just drop the invalidation code since we're hopefully setting a reasonable expiration. 
https://docs.djangoproject.com/en/1.11/topics/cache/

TWLight/resources/views.py
```diff
@@ -42,9 +46,46 @@ def get_queryset(self):
...
+            partners_list = Partner.even_not_available.order_by('company_name')
+        else:
+            partners_list = Partner.objects.order_by('company_name')
+        
+        # Retrieves the short descriptions of all partners from Meta
+        user_language = self.request.LANGUAGE_CODE
+        languages_on_revision_field_short_desc = {}
+        context['short_description'] = {}
+        
+        for every_partner in partners_list:
```
jsnshrmn: We're firing way too much work for each request here. Also, the code here shouldn't be concerned with revision ids or cache invalidation. That's work to do elsewhere, probably `APIPartnerDescriptions`.

TWLight/resources/views.py
```diff
@@ -61,7 +102,93 @@ def get_context_data(self, **kwargs):
                 _("This partner is not available. You can see it because you "
                     "are a staff member, but it is not visible to non-staff "
                     "users."))
-
+
```
jsnshrmn: Same as above and because this logic isn't factored out, you've had to write almost the same code twice.

TWLight/settings/base.py
```diff
@@ -357,4 +357,4 @@ def get_languages_from_locale_subdirectories(dir):
 # REQUEST_LOG_IP. There is not a way to get semi-granular user tracking (such
 # as tracking only authenticated vs anonymous users).
 REQUEST_LOG_IP = False
-REQUEST_LOG_USER = False
+REQUEST_LOG_USER = False
```
jsnshrmn: Why are you messing this EOF?

TWLight/resources/migrations/0062_auto_20190214_1140.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
```
jsnshrmn: no need to include auto generated migrations, it's best if you leave them out to avoid migration conflicts. Not a problem to have it here for now if that's convenient for you, unless we have to rebase and it blows up.

TWLight/resources/models.py
```diff
@@ -200,12 +200,18 @@ class Meta:
         # Translators: In the administrator interface, this text is help text for a field where staff can provide a description of a partner's available resources.
         help_text=_("Optional short description of this partner's resources."))
 
+    short_description_last_revision_ids = models.TextField(blank=True, null=True, editable=False)
```
jsnshrmn: this is a style nitpick, but field names should be singular rather than plural

TWLight/resources/models.py
```diff
@@ -200,12 +200,18 @@ class Meta:
...
     description = models.TextField("long description", blank=True,
         # Translators: In the administrator interface, this text is help text for a field where staff can provide a long description of a partner's available resources.
         help_text=_("Optional detailed description in addition to the short "
         "description such as collections, instructions, notes, special "
         "requirements, alternate access options, unique features, citations notes."))
-        
+    
+    long_description_available = models.BooleanField(default=False)
+    
+    long_description_last_revision_ids = models.TextField(blank=True, null=True, editable=False)
```
jsnshrmn: same as above

TWLight/resources/models.py
```diff
@@ -386,6 +384,10 @@ class Meta:
         # Translators: In the administrator interface, this text is help text for a field where staff can add a description of a collection of resources.
         help_text=_("Optional description of this stream's resources."))
 
+    description_available = models.BooleanField(default=False)
+
+    description_last_revision_ids = models.TextField(blank=True, null=True, editable=False)
```
jsnshrmn: same as above

TWLight/resources/models.py
```diff
@@ -492,4 +489,4 @@ class Meta:
     
     tutorial_video_url = models.URLField(blank=True, null=True,
         # Translators: In the administrator interface, this text is help text for a field where staff can provide links to help videos (if any) for a partner.
-        help_text=_("URL of a video tutorial."))
+        help_text=_("URL of a video tutorial."))
```
jsnshrmn: Why mess with this EOF?

TWLight/resources/tasks.py
```diff
@@ -0,0 +1,26 @@
+from django.core.cache import cache
```
jsnshrmn: So the idea I was trying to get across originally about cache invalidation wasn't that deleting items from the cache needed to be done asynchronously. cache deletion happens in microseconds. I was trying to communicate that we might need to move any expensive queries/network calls required for *cache invalidation logic* out to be asynchronous. However if doing the requests for our cache invalidation logic turns out to be just as expensive as requesting the content, we should replace our concept of invalidation with updating. We could set our ttl to be forever (so that we always have something in the cache) and then go out and fetch a fresh value asynchronously after a reasonable amount of time has passed.

TWLight/resources/templates/resources/partner_tile.html
```diff
@@ -61,7 +60,7 @@
       </div>
     </div>
     <hr />
-    {% if partner.short_description %}
+    {% if short_description|get_desc:partner.pk %}
```
jsnshrmn: style nitpick: put a space around your pipes.

jsnshrmn:
@uyscuti-wiki @Samwalton9 do we want to rebase and complete this before the [python3 migration](https://github.com/WikipediaLibrary/TWLight/milestone/1)? Or do we want to shelve it for later. Happy to plug in the async code after the rebase.

uyscuti-wiki:
If hacking at this wouldn't be too time consuming, I'd say go for it. If not, this can take a back seat for now. Although I'd love to see this code in action in production.

jsnshrmn:
Maybe you could resolve the conflicts then?
`TWLight/resources/templates/resources/partner_users.html`
looks particularly like it could use the author's eyes?

uyscuti-wiki:
Sure thing. Will resolve the conflicts tomorrow.

uyscuti-wiki:
Conflicts are resolved

PR #226, opened 2019-02-08
Don't allow duplicate applications via /apply

Chinmay-Gurjar:
Solved issue of having duplicate applications through /apply
This is with reference to issue T207366.

### DIFF
--- a/TWLight/applications/views.py
+++ b/TWLight/applications/views.py
@@ -115,16 +115,16 @@ class RequestApplicationView(EditorsOnly, ToURequired, EmailRequired, FormView):
         field_order = []
 	open_apps = Application.objects.filter(
 		editor = self.request.user.editor,
-		status__in = (Application.SENT, Application.QUESTION, Application.PENDING)
+		status__in = (Application.SENT, Application.QUESTION, Application.PENDING, Application.APPROVED)
 	)
-	open_apps_list = []
+	open_apps_partners = []
 	for i in open_apps:
-	   open_apps_list.append(i.partner.company_name)
+	   open_apps_partners.append(i.partner.company_name)
         for partner in Partner.objects.all().order_by('company_name'):
             # We cannot just use the partner ID as the field name; Django won't
             # be able to find the resultant data.
             # http://stackoverflow.com/a/8289048
-	    if partner.company_name not in open_apps_list:
+	    if partner.company_name not in open_apps_partners:
                 field_name = 'partner_{id}'.format(id=partner.id)
                 fields[field_name] = forms.BooleanField(
                     label=partner.company_name,
--- a/TWLight/applications/views.py
+++ b/TWLight/applications/views.py
@@ -113,19 +113,27 @@ class RequestApplicationView(EditorsOnly, ToURequired, EmailRequired, FormView):
         """
         fields = {}
         field_order = []
+	open_apps = Application.objects.filter(
+		editor = self.request.user.editor,
+		status__in = (Application.SENT, Application.QUESTION, Application.PENDING)
+	)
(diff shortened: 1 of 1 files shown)


TWLight/applications/views.py
```diff
@@ -113,19 +113,27 @@ def get_form_class(self):
         """
         fields = {}
         field_order = []
+	open_apps = Application.objects.filter(
+		editor = self.request.user.editor,
+		status__in = (Application.SENT, Application.QUESTION, Application.PENDING)
```
Samwalton9: This should also include the `Application.APPROVED` status.

TWLight/applications/views.py
```diff
@@ -113,19 +113,27 @@ def get_form_class(self):
...
+	)
+	open_apps_list = []
```
Samwalton9: `open_apps_list` is a slightly confusing name for this variable, something like `open_app_partners` would be clearer when reading the code.

Chinmay-Gurjar: @Samwalton9 I have made the suggested changes, please review them.

Samwalton9:
This looks great to me! @Chinmay-Gurjar if you're interested, it would be great to follow this PR up with another that adds a test for this functionality!

Otherwise, we still have a bunch of open tasks on Phabricator: https://phabricator.wikimedia.org/project/board/2765/query/rCNFr9ipmEnt/ :)

jsnshrmn:
The build failed out after merge to master. I just restarted it to make sure it wasn't a fluke.
https://travis-ci.org/WikipediaLibrary/TWLight/builds/491589720

jsnshrmn:
and a fluke it was. The build is now passing, meaning that this change will now wend its way into production. Thanks for the contribution!

I do have thoughts about this PR:

#### On style:
For clarity, we try to avoid arbitrary local variable names in our blocks like `for i in open_apps:`, instead we try to be consistent with data structure we're traversing and adjust plurality `for open_app in open_apps:`
Even better, you could have skipped that assignment altogether since you only use it once, and instead followed the style used to traverse the Partner object queryset adjacent to this code. something like:
```
open_apps_partners = []

for application in Application.objects.filter(
    editor = self.request.user.editor,
    status__in = (Application.SENT, Application.QUESTION, Application.PENDING, Application.APPROVED)
):
    open_apps_partners.append(application.partner.company_name)
```

Also, this code should have at least some comments describing what we're doing with partners based on whether or not they have open apps and why (i.e. a quick discussion of the app duplication problem that it's solving).

#### On logic:
Currently, your queryset is acting as a blacklist for `if partner.company_name not in open_apps_partners:`
You can negate matches in querysets so that your queryset actually represents the set of objects that you care about
https://docs.djangoproject.com/en/1.11/ref/models/querysets/#exclude

Additionally, you can filter a queryset based on related objects

By cleaning up your query logic, you can eliminate some extraneous variable assignment and traversal that makes it harder to read than it needs to be. You could end up with something like this:
```
# We exclude partners for whom this editor already has valid applications on file.
for partner in Partner.objects.exclude(
    applications__editor = self.request.user.editor,
    applications__status__in = (Application.SENT, Application.QUESTION, Application.PENDING, Application.APPROVED)
).order_by('company_name')::
    partners_no_open_apps.append(application.partner.company_name)
```

These suggestions are all untested, and the last one contains some hand waiving. If you had trouble collapsing this down into a single queryset and `for` block, you could still reduce your db calls via select_related
https://docs.djangoproject.com/en/1.11/ref/models/querysets/#select-related

And of course, as @Samwalton9 mentioned, tests are always welcome with code! Thanks!

PR #237, opened 2019-03-11
account_email should be an EmailField

Samwalton9:
Nikki noticed an issue with many users skimming over the instructions to pre-register and simply entering "None" for `account_email`. Hopefully enforcing email formatting through `EmailField` form validation will improve this - it probably shouldn't have been a `CharField` to begin with.


Samwalton9:
For completeness, it turns out 'None' was `None` - i.e. these were renewals on apps that didn't have this data in the first place. At any rate this seems a sensible change!

PR #241, opened 2019-03-20
Handle server error

Jain-Aditya:
Gracefully handle server errors on /applications/apply/<random_number>

Phabricator task: https://phabricator.wikimedia.org/T215246


Jain-Aditya:
Hi @Samwalton9 It would be great if you could review this PR. :)

PR #242, opened 2019-03-20
Access code & authorization bug fixes

Samwalton9:
I started working on automatically sending access codes (https://phabricator.wikimedia.org/T206627) and ran into multiple bugs in the access code and authorization code.

### DIFF
--- a/TWLight/applications/models.py
+++ b/TWLight/applications/models.py
@@ -281,20 +281,19 @@ class Application(models.Model):
 
 # IMPORTANT: pre_save is not sent by Queryset.update(), so *none of this
 # behavior will happen on if you update() an Application queryset*.
-# That is sometimes okay (for instance, if you are turning an APPROVED app into
-# a SENT app, since days_open and date_closed were set on approval), but it is
-# not always okay. Errors caused by days_open not existing when expected to
+# That is sometimes okay, but it is not always okay.
+# Errors caused by days_open not existing when expected to
 # exist can show up in weird parts of the application (for example, template
 # rendering failing when get_num_days_open returns None but its output is
 # passed to a template filter that needs an integer).
 @receiver(pre_save, sender=Application)
 def update_app_status_on_save(sender, instance, **kwargs):
 
-    # Make sure not using a mix of dates and dateimes
+    # Make sure not using a mix of dates and datetimes
     if instance.date_created and isinstance(instance.date_created, datetime):
         instance.date_created = instance.date_created.date()
 
-    # Make sure not using a mix of dates and dateimes
+    # Make sure not using a mix of dates and datetimes
     if instance.date_closed and isinstance(instance.date_closed, datetime):
         instance.date_closed = instance.date_closed.date()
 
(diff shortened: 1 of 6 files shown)


Samwalton9:
Still to do:

- Set authorization object correctly when sending access codes - none is set right now, the code is still attempting to connect application objects
- Fix example data for authorizations, including:
  - https://github.com/WikipediaLibrary/TWLight/blob/master/TWLight/users/models.py#L439
  - Multiple authorizations are created from the same user to the same partner, which shouldn't be able to happen
- Add more tests. The first point above especially should have been a test already.

TWLight/applications/models.py
```diff
@@ -319,16 +318,24 @@ def update_app_status_on_save(sender, instance, **kwargs):
...
             instance.days_open = 0
 
-# Authorize editor to access resource after an application is saved as approved
-# or sent. We need to use post_revision_commit from django-reversion to get the
-# **current** revision user. Just listening to the post_save signal will grab
-# the user from the previously saved version.
+# Authorize editor to access resource after an application is saved as sent.
 @receiver(post_save, sender=Application)
 def post_revision_commit(sender, instance, **kwargs):
-    if reversion.is_active() and (instance.status == 2 or instance.status == 4) and not instance.imported:
```
jsnshrmn: I remember now that I wasn't sure which state we wanted to key off of, because the authorization is a logical step that should be triggered by the approval, and be followed by communication about the approval, right?

The check for an existing authorization is definitely a thing that needs to be here, so thanks for that!

TWLight/applications/models.py
```diff
@@ -319,16 +318,24 @@ def update_app_status_on_save(sender, instance, **kwargs):
...
+        existing_authorization = Authorization.objects.filter(
+            authorized_user=instance.user,
+            partner=instance.partner,
+            stream=instance.specific_stream).exists()
+    else:
+        existing_authorization = Authorization.objects.filter(
+            authorized_user=instance.user,
+            partner=instance.partner).exists()
+
+    if instance.status == Application.SENT and not instance.imported: #This work is so close to being done. Final hurdle is that revisions don't get saved with coordinator user when marking sent anymore.
```
jsnshrmn: this looks way better than what I was doing. Is the comment on this line still relevant?

TWLight/applications/models.py
jsnshrmn: Basically, I *think* we need to be creating auths from Approved status instead of sent.

TWLight/applications/models.py
Samwalton9: Oh, nope - this is outdated, I fixed it on line 338, switching from using `reversion` to going direct to `sent_by` which we already have at this point.

TWLight/applications/models.py
jsnshrmn: Per our conversation, we've really morphed our usage of `SENT` to the point where semantically it's something more akin to `FINALIZED`.

TWLight/applications/models.py
Samwalton9: In fact it looks like I remembered to remove it in https://github.com/WikipediaLibrary/TWLight/pull/246/commits/9dc9c9e6bd1027d23d365793c411c8a36206a952 :D

jsnshrmn: works for me!

PR #243, opened 2019-03-21
use include_invalid instead of objects

Jain-Aditya:
Phabricator task:
https://phabricator.wikimedia.org/T215782


Jain-Aditya:
@Samwalton9 Could you please review this?

Samwalton9:
Looks good, tested locally and working as intended - browsing to an invalid application now doesn't server error.

PR #246, opened 2019-03-28
Access code emails

Samwalton9:
This branches from https://github.com/WikipediaLibrary/TWLight/pull/242 so review that first.

### DIFF
--- a/TWLight/applications/models.py
+++ b/TWLight/applications/models.py
@@ -281,20 +281,19 @@ class Application(models.Model):
 
 # IMPORTANT: pre_save is not sent by Queryset.update(), so *none of this
 # behavior will happen on if you update() an Application queryset*.
-# That is sometimes okay (for instance, if you are turning an APPROVED app into
-# a SENT app, since days_open and date_closed were set on approval), but it is
-# not always okay. Errors caused by days_open not existing when expected to
+# That is sometimes okay, but it is not always okay.
+# Errors caused by days_open not existing when expected to
 # exist can show up in weird parts of the application (for example, template
 # rendering failing when get_num_days_open returns None but its output is
 # passed to a template filter that needs an integer).
 @receiver(pre_save, sender=Application)
 def update_app_status_on_save(sender, instance, **kwargs):
 
-    # Make sure not using a mix of dates and dateimes
+    # Make sure not using a mix of dates and datetimes
     if instance.date_created and isinstance(instance.date_created, datetime):
         instance.date_created = instance.date_created.date()
 
-    # Make sure not using a mix of dates and dateimes
+    # Make sure not using a mix of dates and datetimes
     if instance.date_closed and isinstance(instance.date_closed, datetime):
         instance.date_closed = instance.date_closed.date()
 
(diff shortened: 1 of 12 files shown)


Samwalton9:
Hm. I figured Github would realise those commits were now in and bring this down to the two additional commits. Let me know if this needs rebasing.

TWLight/applications/models.py
```diff
@@ -319,16 +318,24 @@ def update_app_status_on_save(sender, instance, **kwargs):
...
-# or sent. We need to use post_revision_commit from django-reversion to get the
-# **current** revision user. Just listening to the post_save signal will grab
-# the user from the previously saved version.
+# Authorize editor to access resource after an application is saved as sent.
 @receiver(post_save, sender=Application)
 def post_revision_commit(sender, instance, **kwargs):
-    if reversion.is_active() and (instance.status == 2 or instance.status == 4) and not instance.imported:
+    # Check if an authorization already exists.
+    if instance.specific_stream:
+        existing_authorization = Authorization.objects.filter(
```
jsnshrmn: It doesn't look like we're actually using this anywhere?

TWLight/applications/models.py
```diff
@@ -319,16 +318,24 @@ def update_app_status_on_save(sender, instance, **kwargs):
...
+            authorized_user=instance.user,
+            partner=instance.partner,
+            stream=instance.specific_stream).exists()
+    else:
+        existing_authorization = Authorization.objects.filter(
+            authorized_user=instance.user,
+            partner=instance.partner).exists()
+
+    if instance.status == Application.SENT and not instance.imported:
```
jsnshrmn: Should we be using the `existing_authorization` in our guard clause? something like this untested code?
```suggestion
    # Create an authorization if this is a newly finalized application for a user that isn't already authrorized.
    if instance.status == Application.SENT and not instance.imported and not existing_authorization:
```

TWLight/emails/tasks.py
```diff
@@ -322,6 +324,33 @@ def update_app_status_on_save(sender, instance, **kwargs):
...
+            base_url = get_current_site(None).domain
+
+            access_code_instructions = instance.partner.access_code_instructions
+            mail_instance = MagicMailBuilder()
+
+            email = mail_instance.access_code_email(instance.authorization.authorized_user.email,
+                {'editor_wp_username': instance.authorization.authorized_user.editor.wp_username,
+                 'partner': instance.partner,
+                 'access_code': instance.code,
+                 'body': access_code_instructions
```
jsnshrmn: I feel like this should probably be `access_code_instructions` instead of `body` for legibility.
```suggestion
                 'access_code_instructions': access_code_instructions
```

TWLight/emails/templates/emails/access_code_email-body-html.html
```diff
@@ -0,0 +1,9 @@
+<html>
+<body>
+  <p>Your approved application to {{ partner }} has now been finalised. Your access code or login details can be found below:</p>
+
+  <p><b>{{ access_code }}</b></p>
+
+  <p>{{ body }}</p>
```
jsnshrmn: I feel like this should probably be `access_code_instructions` instead of `body` for legibility.
```suggestion
  <p>{{ access_code_instructions }}</p>
```

TWLight/emails/templates/emails/access_code_email-body-text.html
```diff
@@ -0,0 +1,5 @@
+Your approved application to {{ partner }} has now been finalised. Your access code or login details can be found below:
+
+{{ access_code }}
+
+{{ body }}
```
jsnshrmn: I feel like this should probably be `access_code_instructions` instead of `body` for legibility.
```suggestion
{{ access_code_instructions }}
```

TWLight/applications/models.py
Samwalton9: Yep, great catch, I wrote the check but forgot to actually use it. This prompted me to give a bit more thought to how authorizations work in relation to renewals. We weren't really handling that properly before, so I've added some logic for updating an existing authorization if one exists and a related test.

Samwalton9: It might seem a little superfluous right now to edit the object, all that's likely to change is the authorizer, so I also added a line for updating expiry dates, which will probably happen here for proxy.

jsnshrmn:
correctly-rendering diff:
https://github.com/WikipediaLibrary/TWLight/compare/master...Samwalton9:access-code-emails

PR #250, opened 2019-04-02
Standalone Authorizations view in users/admin.py

jsnshrmn:
It turns out that it would be really helpful to have a standalone, searchable admin view of authorizations.
I left this out initially because, obnoxiously, when this is added we end up with two "user" apps in admin because of our custom user model. We could concatenate those either with a custom admin template, or maybe [model-admin-reorder](https://django-modeladmin-reorder.readthedocs.io/en/latest/).


Samwalton9:
Looks good to me!

PR #260, opened 2019-04-20
Track number of available accounts per partner, and don't allow more than that to be distributed (T156807)

uyscuti-wiki:
for partners with authorisation method as 'proxy'


TWLight/applications/views.py
```diff
@@ -767,6 +767,34 @@ class EvaluateApplicationView(NotDeleted, CoordinatorOrSelf, ToURequired, Update
     success_url = reverse_lazy('applications:list')
 
     def form_valid(self, form):
+        app = self.object
+        status = form.cleaned_data['status']
+        if app.partner.authorization_method == Partner.PROXY:
+            total_accounts_available_for_distribution = None
+            if app.specific_stream is not None:
+                if app.specific_stream.accounts_available is not None:
+                    total_apps_approved_or_sent = User.objects.filter(
```
Samwalton9: We do this object filtering a few times, both here from this PR and also in other views (partner detail?). Perhaps we should move it to a helper function that takes an application or partner and returns total apps approved or sent?

TWLight/applications/views.py
```diff
@@ -767,6 +767,34 @@ class EvaluateApplicationView(NotDeleted, CoordinatorOrSelf, ToURequired, Update
...
+                    total_accounts_available = app.specific_stream.accounts_available
+                    total_accounts_available_for_distribution = total_accounts_available - total_apps_approved_or_sent
+            elif app.partner.accounts_available is not None:
+                total_apps_approved_or_sent = User.objects.filter(
+                                      editor__applications__partner=app.partner,
+                                      editor__applications__status__in=(Application.APPROVED, Application.SENT),
+                                      editor__applications__specific_stream=None).count()
+                total_accounts_available_for_distribution = app.partner.accounts_available - total_apps_approved_or_sent
+
+            if int(status) == Application.APPROVED:
```
Samwalton9: Should this check be moved up to alongside the proxy check above? It seems like we do a lot of work before checking if the application was approved. If we decline a proxy application I think we would still do all the above calculations, and only realise we didn't need them here, which we should avoid.

TWLight/applications/views.py
```diff
@@ -767,6 +767,34 @@ class EvaluateApplicationView(NotDeleted, CoordinatorOrSelf, ToURequired, Update
...
+                try:
+                    assert app.partner.status != Partner.WAITLIST and (total_accounts_available_for_distribution > 0 if total_accounts_available_for_distribution else True)
+                except AssertionError:
```
Samwalton9: Why `assert` and catch the AssertionError, rather than a simple `if`?

TWLight/applications/views.py
```diff
@@ -841,19 +886,55 @@ def post(self, request, *args, **kwargs):
...
-            app.status = status
-            app.save()
-        #Translators: After a coordinator has changed the status of a number of applications, this message appears.
-        messages.add_message(request, messages.SUCCESS,
-            _('Batch update successful.'))
+            if app.partner.authorization_method == Partner.PROXY:
+                total_accounts_available_for_distribution = None
+                if app.specific_stream is not None:
+                    if app.specific_stream.accounts_available is not None:
+                        total_apps_approved_or_sent = User.objects.filter(
```
Samwalton9: Seems quite computationally expensive to do this check for every application in the list. Could we calculate the number once before looping, and then track how many we have left with a simple variable, rather than re-filtering every loop?

TWLight/applications/views.py
```diff
@@ -841,19 +886,55 @@ def post(self, request, *args, **kwargs):
...
+                    total_apps_approved_or_sent = User.objects.filter(
+                                          editor__applications__partner=app.partner,
+                                          editor__applications__status__in=(Application.APPROVED, Application.SENT),
+                                          editor__applications__specific_stream=None).count()
+                    total_accounts_available_for_distribution = app.partner.accounts_available - total_apps_approved_or_sent
+
+                if int(status) == Application.APPROVED:
+                    try:
+                        assert app.partner.status != Partner.WAITLIST and (total_accounts_available_for_distribution > 0 if total_accounts_available_for_distribution else True)
+                        batch_update_success.append(app_pk)
```
Samwalton9: This is a nice way of approving as many as possible before hitting the limit, but I wonder if we'd be better denying the batch outright so that the coordinator can decide which of the applications they want to approve, so that they can prioritise. I'll ask Nikki on Slack.

uyscuti-wiki:
> Looking good overall! Left a few comments throughout.
> 
> One larger issue is that this code should be checking how many active authorizations we have, not how many applications we've marked approved or sent. For proxy Authorization objects (with their expiry dates) will give us a direct measure of how many active accounts we have, so we should use them to find out how many accounts we have left to distribute.

Question: does this mean we reset the current `accounts_available` field to the then available accounts for distribution via proxy?

Samwalton9:
It's probably easiest to keep parity with the current workflow. That is, `accounts_available` doesn't change and we simply compare the number of active authorizations with it.

uyscuti-wiki:
Does that mean all applications (both historical and recent) for a particular partner with authorisation method either as proxy or access code have authorisation objects associated with them? 

For instance, Gale has 200 accounts available, and it's [active authorisations](https://wikipedialibrary.wmflabs.org/admin/users/authorization/?o=3) (assuming this is what you refer to as 'active') seem to only be two?

Samwalton9:
Not right now - we haven't yet had a need to go back and create authorization objects for historical applications, but we're creating them going forward. When I say 'active' authorizations I guess I really mean their expiry date is prior to today. Perhaps we should have a property on authorization objects that checks that, seems like a useful function to have centralised.

TWLight/applications/views.py
uyscuti-wiki: no particular reason

TWLight/applications/views.py
uyscuti-wiki: Correct me if I'm wrong, to compute it once before the loop is to assume every single approval in a batch is for the same partner, is it not?

uyscuti-wiki:
Is [this](https://github.com/WikipediaLibrary/TWLight/pull/260/files#diff-8e6b10cdc0483601766b26ac17dc3a89R160) what you were hoping for?

TWLight/applications/views.py
uyscuti-wiki: we could, however, filter out duplicates in the batch and query once for every unique partner/collection which I guess would marginally decrease overhead.

uyscuti-wiki: well, that would make the whole thing pointless because we then would have to do a redundant `Application.objects.get(pk=app_pk)`.

uyscuti-wiki: figured out a [workaround](https://github.com/WikipediaLibrary/TWLight/pull/260/files#diff-dba635dfa3cab82282374d6f91d3b958R878)

Samwalton9: I did forget that we could be batch-editing across multiple partners :)

TWLight/applications/views.py
Samwalton9: Per Nikki's feedback on Slack, it would be preferable to take no actions so that the coordinator can decide which applications to approve.

TWLight/applications/views.py
uyscuti-wiki: > Per Nikki's feedback on Slack, it would be preferable to take no actions so that the coordinator can decide which applications to approve.

This is going to make the workaround redundant.

uyscuti-wiki:
How does this look now?

TWLight/resources/tests.py
```diff
@@ -695,3 +698,77 @@ def test_csv_formatting(self):
...
+            agreement_with_terms_of_use=True)
+
+        self.coordinator = UserFactory(username='coordinator')
+        self.coordinator.set_password('coordinator')
+        coordinators = get_coordinators()
+        coordinators.user_set.add(self.coordinator)
+        self.coordinator.userprofile.terms_of_use = True
+        self.coordinator.userprofile.save()
+
+        for _ in range(2):
```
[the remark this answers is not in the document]
Samwalton9: That said, you should double check this, because if it's only creating one authorization I don't see why the test would pass. Perhaps I just misunderstand how this works :)

bin/twlight_unwaitlist.sh
```diff
@@ -0,0 +1,18 @@
...
+# Environment variables may not be loaded under all conditions.
+if [ -z "${TWLIGHT_HOME}" ]
+then
+    source /etc/environment
+fi
+
+# Load virtual environment
+if source ${TWLIGHT_HOME}/bin/virtualenv_activate.sh
+then
+    echo "Probing for waitlisted partners with atleast one active authorization"
```
Samwalton9: Could you reword this? It's a little unclear. We're looking for partners with at least one available account, not one active authorization, right?

TWLight/applications/tests.py
```diff
@@ -2533,138 +2631,25 @@ def test_under_discussion_signal(self):
 
         self.assertEqual(self.application.status, Application.QUESTION)
 
-    def test_immediately_sent_collection(self):
```
Samwalton9: Why were these tests removed? Just a rebase conflict?

TWLight/applications/tests.py
```diff
@@ -2392,6 +2393,103 @@ def test_sets_status(self):
         self.assertEqual(self.application.status, Application.APPROVED)
 
 
+    def test_sets_status_sent_for_proxy_partner(self):
```
Samwalton9: Could you add comments to these tests explaining what they're testing?

TWLight/applications/signals.py
```diff
@@ -17,4 +23,23 @@ def under_discussion(sender, comment, request, **kwargs):
 
     if application_object.status == Application.PENDING and application_object.editor.user.username != request.user.username:
         application_object.status = Application.QUESTION
-        application_object.save()
\ No newline at end of file
+        application_object.save()
+
+
+@receiver(no_more_accounts)
+def set_partner_status(sender, **kwargs):
```
Samwalton9: Final thought: Is there a reason this is a separate and manually triggered signal, rather than having this be part of the `post_save` trigger for Applications? My initial thought with signals was that was where it would happen so that _however_ we set an application approved, it would always trigger, rather than requiring a signal be sent manually.

uyscuti-wiki: I think signals.py is a more appropriate place to do this since we are trying to make some changes from `applications` that would implicate the `resources` application, I could be wrong though. I also like the manual trigger of signal for readability purposes.

Samwalton9: >we are trying to make some changes from applications that would implicate the resources application

That's not inherently an issue - in the current `post_save` we're creating Authorization objects, for example: https://github.com/WikipediaLibrary/TWLight/blob/master/TWLight/applications/models.py#L356

@jsnshrmn do you have thoughts on this? The core issue being triggering a manual signal to set partners to waitlist when approving the final available application vs. having the logic in the application `post_save` signal.

jsnshrmn: I agree with @Samwalton9, but I've also let this sit for quite some time. I'm inclined to go ahead and approve this since the code is working.

PR #269, opened 2019-05-30
Allow users to choose 1, 3, 6 or 12 month account lengths for proxy (T170383)

uyscuti-wiki:
both when applying and renewing


TWLight/applications/forms.py
```diff
@@ -319,3 +319,38 @@ def __init__(self, user=None, *args, **kwargs):
...
+            raise
+        self.field_params = kwargs.pop('field_params')
+        super(RenewalForm, self).__init__(*args, **kwargs)
+        self.helper = FormHelper(self)
+        self.helper.layout = Layout()
+        # add translator comment
+        fieldset = Fieldset(_('Renewal confirmation'))
+        if 'account_email' in self.field_params:
+            self.fields['account_email'] = forms.EmailField()
+            self.fields['account_email'].label = _('The email for your account on the partner\'s website')
```
Samwalton9: Bonus points if we can pre-fill this with the `account_email` data from the previous application!

TWLight/applications/forms.py
```diff
@@ -319,3 +319,38 @@ def __init__(self, user=None, *args, **kwargs):
...
+            fieldset.append('account_email')
+        if 'proxy_account_length' in self.field_params:
+            self.fields['proxy_account_length'] = forms.ChoiceField(choices=Application.PROXY_ACCOUNT_LENGTH_CHOICES,)
+            self.fields['proxy_account_length'].label = _('The number of months you wish to have this access for')
```
Samwalton9: We should probably say '...before renewal is required' or something at the end, so as not to imply that it's a one-off thing.

TWLight/applications/models.py
```diff
@@ -102,6 +103,13 @@ class Meta:
     agreement_with_terms_of_use = models.BooleanField(default=False)
     account_email = models.EmailField(blank=True, null=True)
 
+    PROXY_ACCOUNT_LENGTH_CHOICES = ((1, '1 month'), (3, '3 months'), (6, '6 months'), (12, '12 months'))
```
Samwalton9: Since we already said "number of months" in the form label, this can probably just be a list of numbers. Will save us translating each of these strings.

TWLight/resources/models.py
```diff
@@ -328,6 +328,10 @@ class Meta:
         # Translators: In the administrator interface, this text is help text for a check box where staff can select whether users must first register at the organisation's website before finishing their application.
         help_text=_("Mark as true if this partner requires applicants to have "
                     "already signed up at the partner website."))
+    proxy_account_length = models.BooleanField(default=False,
```
Samwalton9: We maybe don't need this - I think ideally we would have this selector for all proxy partners.

TWLight/applications/models.py
```diff
@@ -102,6 +103,13 @@ class Meta:
     agreement_with_terms_of_use = models.BooleanField(default=False)
     account_email = models.EmailField(blank=True, null=True)
 
+    PROXY_ACCOUNT_LENGTH_CHOICES = ((12, '12 months'), (6, '6 months'), (3, '3 months'), (1, '1 month'))
```
Samwalton9: Needs flagging for translation.

TWLight/applications/models.py
```diff
@@ -141,6 +149,7 @@ def renew(self):
         successful and None otherwise.
         """
         if not self.is_renewable:
+            logger.info('None')
```
Samwalton9: Was this just for testing? If not, needs a more descriptive message :)

TWLight/applications/models.py
```diff
@@ -403,11 +413,18 @@ def post_revision_commit(sender, instance, **kwargs):
         authorization.authorizer = authorizer
         authorization.partner = instance.partner
 
-        # If this is a proxy partner, set (or reset) the expiry date
+        # If this is a proxy partner, and the proxy_account_length
+        # field is set to false, set (or reset) the expiry date
         # to one year from now
-        if instance.partner.authorization_method == Partner.PROXY:
+        if instance.partner.authorization_method == Partner.PROXY and (instance.partner.proxy_account_length is False or instance.proxy_account_length is None):
```
Samwalton9: If we've set up the validation for admin correctly, should it ever be the case that `instance.partner.proxy_account_length is False` while authorization method is Proxy?

uyscuti-wiki: I forgot about this part and I think we agreed we should have a validation because that's never the case. I'll make sure to add it.

TWLight/static/css/local-rtl.css
```diff
@@ -250,6 +250,10 @@ ul.timeline-home li {
 MISC
 --------------------------------------------------------------------------*/
 
+div.checkbox#div_id_cc {
```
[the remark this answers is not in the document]
uyscuti-wiki: [this](https://github.com/WikipediaLibrary/TWLight/commit/d1ea31b4fd325c7e437e588b0cee94c6370bcb17#diff-cf70181249768f33984ea54eceff88c8)

TWLight/applications/templates/applications/application_evaluation.html
```diff
@@ -70,6 +70,18 @@ <h2>{% trans 'Application' %}</h2>
...
     </div>
 
+    {% if object.proxy_account_length %}
+      <div class="col-xs-12 col-sm-4">
+        {% comment %} Translators: This labels the section of the application showing the length of the account (expiry) requested by the user.{% endcomment %}
+        <strong>{% trans 'Account length' %}</strong>
+      </div>
+      <div class="col-xs-12 col-sm-8">
+        {{ object.proxy_account_length }}
+        {% comment %} Translators: This adds the word 'month(s)' next to the actual number of months.{% endcomment %}
```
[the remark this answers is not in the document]
uyscuti-wiki: [this](https://github.com/WikipediaLibrary/TWLight/pull/130/files#diff-7ba5a89d21b1584637ddf7290147b611R232) seems to work just fine, any particular reason why you think this wouldn't work here?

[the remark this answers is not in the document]
Samwalton9: Ah, maybe it's fine then - nevermind!

Samwalton9:
Don't disagree with making the field more broadly applicable, but I don't think it should default to 1 year - that was the situation in the past and we removed it because some manual-access accounts last indefinitely.

jsnshrmn:
Right, right. In that case, we should probably make sure the authorization reporting code counts auths with no end date.

Samwalton9:
re: https://github.com/WikipediaLibrary/TWLight/pull/282? Good point. Not an issue for now because we only check authorizations for proxy partners, otherwise reporting sent applications. Ignored the nuance of non-proxy partners with useful authorization objects.

uyscuti-wiki:
Refactored to `requested_access_duration`

PR #270, opened 2019-06-11
Docker

jsnshrmn:
This will be a breaking change that requires developers to throw away their vagrant environments and for us to redeploy servers. Not merged yet, but I wanted to give contributors a heads up before it happens.
@Samwalton9 @uyscuti-wiki @Jain-Aditya @Chinmay-Gurjar @arcayn


README.md
```diff
@@ -2,6 +2,37 @@
 
 The Wikipedia Library Card Platform app, available at http://wikipedialibrary.wmflabs.org.
 
+## Quick Installation for Developers
+
+- Get Docker and Docker Compose
+- Fire up an empty TWLight instance `docker-compose build && docker-compose up`
+- Run migrations `docker-compose exec twlight /app/bin/virtualenv_migrate.sh`
```
Samwalton9: Is there a reason we wouldn't want to always run migrations in the Dockerfile?

jsnshrmn: - we don't have a database at build time (when the dockerfile is in play), but on the servers, I am automatically running the migration upon having the image come up.
- there are situations in which you really don't want automatic migration generation in dev, like when you are working on code that requires a custom migration. This occasionally bit me in the vagrant environment, and I'd have to run through some hoops to avoid the migrations.

Having said that, we should probably have feature parity with the vagrant environment in terms of what you get when you bring the environment up. I'm kind of an edge case.

Samwalton9: Gotcha - that totally makes sense, and one line isn't really an issue, I was just curious.

PR #272, opened 2019-06-13
Add PyCharm + Docker setup instructions

Samwalton9:


docs/developer.md
```diff
@@ -75,3 +73,24 @@ All four places referenced there must be updated:
...
+
+## PyCharm setup
+
+This project can be set up via PyCharm using its support for Docker. Wikimedia developers can get free access to PyCharm Professional (required for Docker support) - please contact The Wikipedia Library for instructions.
+
+### Process
+
+1. After installing Docker, ensure it's using Linux Containers, and in settings, enable the 'Expose daemon on tcp://localhost:2375 without TLS' option in settings.
+2. In PyCharm, open the repository folder, and navigate to Project Settings (File > Settings)
+3. Under Build, Execution, Deployment > Docker, click the + symbol and check the TCP socket option is selected.
```
uyscuti-wiki: and the Engine API URL is `tcp://localhost:2375`

docs/developer.md
```diff
@@ -75,3 +73,24 @@ All four places referenced there must be updated:
...
+4. Navigate to Project > Project Interpreter, click the cog icon, then Add... and select the Docker Compose tab
```
uyscuti-wiki: cog/gear

docs/developer.md
```diff
@@ -75,3 +73,24 @@ All four places referenced there must be updated:
...
+- Configuration file(s) should be set to both `docker-compose.yml` and `docker-compose.override.yml`
+- Service should be set to `twlight`.
+- Set Python interpreter path to /venv/bin/python
+5. It may take some time for the interpreter to finish adding. If successful, you should see a long list of python packages including Django.
+6. Open the Docker tab in PyCharm's bottom bar. With the Docker entry highlighted, click the icon with three green arrows, then `Create docker-compose deployment...`
+- Compose file(s) should again be set to both `docker-compose.yml` and `docker-compose.override.yml`
+- Check the `--build` option
+- Click Apply, then Run
+7. You should see the three containers - `twlight_docker_db_1`, `twlight_docker_twlight_1`, and `twlight_docker_web_1` start up correctly.
```
uyscuti-wiki: you should also see the Library Card platform when you visit `localhost` in your browser.

docs/developer.md
```diff
@@ -75,3 +73,24 @@ All four places referenced there must be updated:
...
+8. Whenever you want to start the Docker containers again, you can simply click the green arrows, then 'Docker'.
```
uyscuti-wiki: Would be worth including two of the potential issues I/we came across and their solutions. Namely, `Cannot start service db:` and `Drive has not been shared`. As an additional tip would be nice to include the process to execute scripts and maybe restart gunicorn.

docs/developer.md
```diff
@@ -2,21 +2,19 @@
...
+These scripts can be run with, for example `docker exec twlight virtualenv_migrate.sh`.
 
 ### Virtual environment
 
 To activate the Python virtual environment for the project, run:
 
-`sudo su www`
-
-`source /var/www/html/TWLight/bin/virtualenv_activate.sh`
+`/app/bin/virtualenv_activate.sh`
```
jsnshrmn: Is this a thing that still needs to happen? `docker exec twlight bash` should already bring you into an activated environment.

Samwalton9: Ah, you're right. Removed that section.

Samwalton9:
Done and done :)

PR #281, opened 2019-06-25
Track average authorizations per user

Samwalton9:
https://phabricator.wikimedia.org/T224550


Samwalton9:
@jsnshrmn Would like to be able to report this stat at the end of the week - could you take a look before then?

Samwalton9:
Good point - clarifying with Nikki.

Samwalton9:
Yep we want to exclude expired auths. Will fix that later or tomorrow morning.

PR #289, opened 2019-07-11
Rework 404s on partner CSVs

Samwalton9:
Nothing was breaking here, we were just receiving exception emails for situations where all we're doing is presenting a 404.


Samwalton9:
This is straightforward and saves my inbox so merging straight in.

PR #299, opened 2019-08-22
Get users with missing Editor objects out of authentication limbo

jsnshrmn:
User <=> Editor relationship is now mutable. This has consequences.


Samwalton9:
LGTM! Tested (or worth testing) on Staging?

PR #302, opened 2019-09-19
My Collection page (T230525) & List active authorizations on user pages (T161170) & Allow users to 'return' their proxy-based access (T169201)

uyscuti-wiki:
This PR breaks out 'list applications' which was part of the user profile to a separate page. Also adds a new 'My collection' page along with some minor aesthetic changes to the base template.


TWLight/users/views.py
```diff
@@ -572,3 +573,108 @@ def get(self, request, pk, version, format=None):
...
+                                                             Q(date_expires__lt=today),
+                                                             Q(partner__authorization_method=Partner.EMAIL) |
+                                                             Q(partner__authorization_method=Partner.CODES) |
+                                                             Q(partner__authorization_method=Partner.LINK)
+                                                            ).order_by('partner')
+        for each_authorization in proxy_bundle_authorizations:
+            if each_authorization.date_expires is not None:
+                if each_authorization.date_expires - today < timedelta(days=30):
+                    each_authorization.about_to_expire = True
+                    try:
```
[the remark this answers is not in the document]
Samwalton9: i.e. `Authorization.get_latest_app()`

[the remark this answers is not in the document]
Samwalton9: I can't find where else we're doing this but I distinctly remember needing to do either this or the reverse (get auth for an application) somewhere else.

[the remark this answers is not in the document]
Samwalton9: Ah, I might have been thinking of the reverse at https://github.com/WikipediaLibrary/TWLight/blob/master/TWLight/applications/views.py#L1027

Either way, this is a useful thing to have centralised, especially since you're doing it a few times in this file alone.

TWLight/applications/models.py
```diff
@@ -25,6 +25,14 @@ def get_queryset(self):
         return super(ValidApplicationsManager, self).get_queryset(
             ).exclude(status=Application.INVALID)
 
+
+def get_latest_app(partner, editor):
```
Samwalton9: This can be directly on the Authorization model so that you don't need to pass the partner and editor info in. See e.g. the functions at https://github.com/WikipediaLibrary/TWLight/blob/master/TWLight/users/models.py#L220

Samwalton9: Then you wouldn't need to do https://github.com/WikipediaLibrary/TWLight/blob/5c5797f4b9018cb93ecbda2cb59d9cb810bf387a/TWLight/users/views.py#L620 because you could access that data directly, without needing to loop through each authorization object and adding new data. In fact, the same could be said for expiring soon and expired. If those were functions on the Authorization model we could potentially do away with needing those loops at all?

uyscuti-wiki: importing Application model in `users.model` results in all sorts of errors. Not sure how to proceed.

uyscuti-wiki: is this what is called a 'circular import'? Should I try to import from within the function?

Samwalton9: That issue is avoidable in Django, see https://github.com/WikipediaLibrary/TWLight/blob/28a9ce3f77406c8613596badc3a22072d20018f9/TWLight/resources/models.py#L578.

uyscuti-wiki: sorry, but I don't fully understand what you are trying to ask of me. Establish Many-to-one relationship from Application to Authorization and then tweak the code to set `Authorization.pk` for every application that is being created? If you had something else in your mind, please explain.

Samwalton9: Ah, sorry for the confusion, I'm probably using the wrong wording. I'm referring to adding model methods (see https://django-book.readthedocs.io/en/latest/chapter10.html#model-methods) to the Authorization model so this functionality lives more centrally.

On the user model, for example, we have a `get_absolute_url` method (https://github.com/WikipediaLibrary/TWLight/blob/master/TWLight/users/models.py#L374) which can be called on any user object. We could have `get_latest_app` be a method on the Authorization model so that by default you can always do `authorization_object.get_latest_app()` to find the application associated with an Authorization object.

See the bottom of https://gist.github.com/Samwalton9/aee729e6f4faa8347e0ca99f2f40d6e2 for an (untested) example of what I mean. With `get_latest_app` as a model method directly, you then wouldn't need to loop through all your applications setting that information here, because the function would already be accessible to you directly on the model.

Does that clarify?

uyscuti-wiki: Yep, that clarifies things. Thanks!

uyscuti-wiki:
@Samwalton9 looks reasonable now?

Samwalton9:
I think the information text after clicking X could do with a little expansion to clarify what 'return' means. I'd suggest adding this to the end:

"You will no longer be able to access {partner} resources via the Library Card platform, but can request access again if you change your mind."

uyscuti-wiki:
@jsnshrmn all instances of `authorized_user` is refactored to `user`.

jsnshrmn:
Okay, there is now a `get_access_url()` property of partners and streams that will return the full proxied url for proxied resources, or the target_url for resources without proxy. If there is no target url it returns `None`
So a template to produce an access link to a resource object might look something like:
```
{% if object.get_access_url %}
<a href="{{ object.get_access_url }}">Access Now</a>
{% endif %}
```

uyscuti-wiki:
@jsnshrmn the refactoring doesn't seem to have worked as well as I'd hoped. While everything seems normal in the Authorization admin view, take a look here http://twlight-staging.wmflabs.org/admin/resources/partner/24/change/ - `authorized_user` still exists and expects input, but ignores the input.

jsnshrmn:
You just missed a reference in one of the forms called by admin.

Samwalton9:
Quick question - does this allow renewal of _all_ proxy authorizations, or only those which have renewals available? We'll want the latter & will set most if not all to available. But if partners leave TWL they're sometimes happy for existing accounts to expire naturally rather than forcefully removing them, so we'd want that functionality of having an authorization but not allowing a renewal.

uyscuti-wiki:
@Samwalton9 because we now intercept all renewals, the functionality is already available and renewals are restricted when the `renewal_available` field is unchecked. Does that cover this use case?

Samwalton9:
"renewals are restricted when the renewals_available field is unchecked" This covers what I was asking - we don't want to show the Renew button if they're not available :)

uyscuti-wiki:
Button will be disabled when renewals are not available with a tooltip – gives users a bit more info on why they can't renew.

TWLight/users/views.py
```diff
@@ -579,3 +580,109 @@ def get(self, request, pk, version, format=None):
...
+        try:
+            return Editor.objects.get(pk=self.kwargs['pk'])
+        except Editor.DoesNotExist:
+            raise Http404
+
+    def get_context_data(self, **kwargs):
+        context = super(CollectionUserView, self).get_context_data(**kwargs)
+        editor = self.get_object()
+        today = datetime.date.today()
+        proxy_bundle_authorizations = Authorization.objects.filter(Q(date_expires__gte=today) |
```
jsnshrmn: the indentation here gets ridiculous. We should probably set some project config/hints for pycharm that we keep in version control. Not at all a blocker, just an observation.

PR #313, opened 2019-09-30
Improve test coverage and organization for authorization objects.

jsnshrmn:
currently tests that touch on authorization objects are quite scattered since they sprouted out of a cross-cutting concern. There are definitely bits that need coverage added, but also a reorg of some of the tests would make our coverage of authorizations more legible.


TWLight/tests.py
```diff
@@ -429,22 +408,436 @@ def test_email_required_2(self):
...
+            partner=self.partner1,
+        )
+
+        # Send the application
+        self.client.post(
+            reverse("applications:evaluate", kwargs={"pk": self.app3.pk}),
+            data={"status": Application.SENT},
+            follow=True,
+        )
+        self.app3.refresh_from_db()
```
Samwalton9: Is it reliable to do client POSTs in setUp? Feels like we should make the objects directly for reliability, but I'm not sure.

jsnshrmn: should be totes reliable. I like doing the process in the same number of steps as the real world, e.g. create the pending application, then have an coordinator approve it. Doing it that way creates all of the appropriate revisions of the application. I could switch the over to a factory on the evaluate view instead of calling the client to make it more like the tests that follow. I had a vague recollection that factory requests had some limitations that came into play with the application revisions, but I'll go double-check.

jsnshrmn: Just went through and tested doing factory request posts in setup, and those don't work. The client-based app updates work just fine in setup though.

jsnshrmn: dug into it just a little more to see why, and basically the answer is authentication. I could go through and patch up the factory coordinator and the factory requests, but that's basically exactly what the client is for.

PR #318, opened 2019-10-03
Account for LINK partners and update out-of-date text

uyscuti-wiki:


Samwalton9:
Could you summarise what this PR does? It's quite hard to tell from the Github diffs!

uyscuti-wiki:
The 'Application data' section is wrapped in a condition to check if the Partner.LINK is true. If true, would ignore the section. Also updated the wording for the else case (Bundle/Proxy).

Samwalton9:
Ah, gotcha. Should users even be able to visit the Mark as Sent page for a Link/Proxy/Bundle partner? It shouldn't ever arise that they're able to see a link leading there, right?

If so, seems like we should probably not support navigating to these pages and do away with the conditions altogether. Can chat about this at library card check-in :)

uyscuti-wiki:
It's not totally useless for LINK partners atm. Send instructions _are_ part of this page. But for others, it is true. Unless coordinators go out of their way manually plugging the link to PROXY/BUNDLE partners, they won't end up on this page, and even if they did end up, it's an empty harmless page. Restricting access to this page for LINK/BUNDLE/PROXY and showing send_instructions elsewhere for LINK partners could be low priority task for the future. But for now, this shouldn't break anything and is a simple fix.

Samwalton9:
Absolutely true. Mostly looking to cut down on unnecessary considerations & translations, but agree this seems to be a good quick fix for an outstanding issue. Will defer to Jason for final review :)

PR #319, opened 2019-10-04
document import formats

jsnshrmn:


Samwalton9:
Oh, nice! The only other thing that would be nice to note is which fields are required. I think it's just username and partner ID?

jsnshrmn:
ha! caught me midway through. How does this look?

PR #322, opened 2019-10-04
Don't show Send if application should be instantly finalized (T221851)

Samwalton9:


Samwalton9:
Had to comment out the use of `authorizer` to find an application's authorization because https://phabricator.wikimedia.org/T233508 prevents it from being set correctly. Partner-User(-Stream) should be a unique combination anyway.

PR #330, opened 2019-10-18
WIP proxy approval email

jsnshrmn:
somewhat rejiggers where the static/template-based user instructions live.


jsnshrmn:
I getting some unexpected ezproxy test failures with this locally, so I'm just going to circle back around to it with fresh eyes, or have @Samwalton9 take a look for something silly.

jsnshrmn:
Passes all tests on Travis... Interesting.
I've been hopping around various forks and branches on a single running container today, so it could be that I just need clear out the local cruft.

jsnshrmn:
and really, get_user_instructions() should probably live in resources instead of applications.

TWLight/applications/models.py
```diff
@@ -266,10 +266,25 @@ def get_user_instructions(self):
...
-            return self.partner.user_instructions
+            resource = self.specific_stream
+        elif self.partner:
+            resource = self.partner
+
+        # Fetch instructions from the database if appropriate
+        if resource:
+            if resource.authorization_method in [Partner.CODES, Partner.LINK] and resource.user_instructions:
+                user_instructions = resource.user_instructions
+            elif resource.authorization_method == Partner.PROXY and resource.get_access_url:
```
Samwalton9: If authorization method is PROXY and we don't have an access URL, should we error out instead of sending the 2-4 weeks text email? Doesn't seem that's a situation we should ever want to be in.

Samwalton9: Likewise, should we handle having CODES or LINK but no user instructions here?

jsnshrmn: > 
> 
> If authorization method is PROXY and we don't have an access URL, should we error out instead of sending the 2-4 weeks text email? Doesn't seem that's a situation we should ever want to be in.

We should probably just make `target_url` a required field for all resources. After all, these are 100% web-accessible resources, and that will make lots of ui and logic stuff simpler if we always have a representative url stored there.

Samwalton9: Yep, that makes sense.

jsnshrmn: > Likewise, should we handle having CODES or LINK but no user instructions here?
I spent a little time fiddling with this. The trick about raising errors here is that (in the context of the email messages) the approval/send has already happened even though it can't really work.

Since this is just a getter property on a model, really the correct thing to do here is to get the thing in question if possible, and return nothing if it's not.

We could add some admin validation (like with account length) for CODES and LINK resources to help with new partners and add some validation checks (with helpful messages) in application pre_save to make sure we don't allow apps to be approved if we can't provide access to the editor upon send.

Samwalton9: That makes sense to me!

jsnshrmn: Excellent. In the interest of keeping things moving, let's go ahead and kick that out to a separate issue, since the problem wasn't introduced in this pr.

Samwalton9: Filed https://phabricator.wikimedia.org/T236390

PR #331, opened 2019-10-21
Show renew button for authorizations with empty expiry dates (T236086)

uyscuti-wiki:
https://phabricator.wikimedia.org/T236086

### DIFF
--- a/TWLight/users/templates/users/collection_tile.html
+++ b/TWLight/users/templates/users/collection_tile.html
@@ -68,23 +68,23 @@
       <hr>
     {% endif %}
     <i class="fa fa-info-circle" aria-hidden="true" title="Partner description page"></i> <strong><a href="{% url 'partners:detail' each_authorization.partner.pk %}">{{ each_authorization.partner }}</a></strong>
+    {% if each_authorization.date_expires and each_authorization.about_to_expire or not each_authorization.is_valid or not each_authorization.partner.authorization_method == each_authorization.partner.PROXY and not each_authorization.partner.authorization_method == each_authorization.partner.BUNDLE %}
+      {% if each_authorization.latest_app and not each_authorization.open_app %}
+        {% comment %} Translators: Hovertext for when renewals for a partner is not available and the renew button is disabled. {% endcomment %}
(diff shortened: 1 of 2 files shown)


TWLight/users/templates/users/collection_tile.html
```diff
@@ -68,23 +68,23 @@
...
-              {% trans 'Renew' %}
-            </a>
-          </span>
-          <div class="clearfix"></div>
-        {% endif %}
-      {% endif %}
-      {% if each_authorization.open_app %}
-        {% comment %} Translators: Labels the button that links to the application page itself if the applications is already renewed. {% endcomment %}
-        <a href="{% url 'applications:evaluate' each_authorization.open_app.pk %}" class="btn btn-primary pull-right" role="button">{% trans 'View application' %}</a><div class="clearfix"></div>
+    {% if each_authorization.date_expires and each_authorization.about_to_expire or not each_authorization.is_valid or not each_authorization.partner.authorization_method == each_authorization.partner.PROXY and not each_authorization.partner.authorization_method == each_authorization.partner.BUNDLE %}
```
Samwalton9: This is a _very_ long line - can we simplify it via the model properties at all? For example, could `about_to_expire` include a check for `date_expires`?

uyscuti-wiki: Hmm https://github.com/WikipediaLibrary/TWLight/pull/302/commits/a55e3e5d64ee65392685d81c7ce08dc3be6b88ed#diff-fb728e27df603ec50dac4ccfdb694b8fL513 :)

uyscuti-wiki: happy to add the model back in of course

Samwalton9: Could you elaborate? We removed `is_expired` because we have the opposite function `is_valid`, right?

And actually i just took a quick look and `about_to_expire` already includes `if authorization.date_expires`, so that's an easy win in terms of reducing code here.

uyscuti-wiki: sorry, I was thinking of a different thing. Didn't realise I'd already had a check in `about_to_expire` for `date_expires` – you're right, will remove the redundant check.

PR #333, opened 2019-10-22
fill in missing auth expiration dates upon partner and stream save.

jsnshrmn:


TWLight/users/signals.py
```diff
@@ -1,4 +1,41 @@
...
+    """
+    Updates missing expiration dates upon resource updates.
+    Mostly cribbed from
+    TWLight.applications.models.post_receive_commit
+    and
+    TWLight.users.management.commands.authorization_backfill
+    Could factor out the common code.
+    """
+
+    print(sender)
```
Samwalton9: print!

TWLight/users/signals.py
```diff
@@ -1,4 +1,41 @@
...
+    if sender == Partner:
+        partner = instance
+    elif sender == Stream:
+        partner = instance.partner
+
+    if partner.account_length or partner.requested_access_duration:
```
Samwalton9: `requested_access_duration` is effectively a dummy field for generating the applications in a consistent way. We should probably just check for `.PROXY` to be clearer about what's happening here?

TWLight/users/signals.py
```diff
@@ -1,4 +1,41 @@
...
+        authorizations = Authorization.objects.filter(partner=partner)
```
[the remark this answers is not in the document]
Samwalton9: Oh I've just noticed we check for `not authorization.date_expires` below. Could it make sense to just filter on that in the first place?

TWLight/users/signals.py
```diff
@@ -1,4 +1,41 @@
...
+        for authorization in authorizations:
+            if authorization.is_valid and not authorization.date_expires:
+                if partner.authorization_method == Partner.PROXY and partner.requested_access_duration is True:
```
Samwalton9: In the `.PROXY` case, we only want to give authorizations a 365 day expiry date if we expect the manual account was still active. By this I mean `today < authorization_date + account_length` if we have `account_length`, or all authorizations otherwise. In many cases we have ~100 accounts distributed but know they only lasted 1 year, so everything from over a year ago shouldn't be 'reactivated'.

TWLight/users/signals.py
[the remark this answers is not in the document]
jsnshrmn: Yeah, I didn't do it that way, because I thought there might be situations in which we needed to update existing expiration dates. I meant to pose that question in the pr body, but instead just posted it empty. If we really only care about empty expiration dates, then filtering the query set would definitely be more performant

TWLight/users/signals.py
jsnshrmn: I knew there was some element that I was forgetting. Thanks!

TWLight/users/signals.py
jsnshrmn: 😳

TWLight/users/signals.py
jsnshrmn: not sure what I was thinking here

TWLight/users/signals.py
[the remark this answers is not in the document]
Samwalton9: I think if there's an existing expiration date (only currently the case in a handful of places if I recall) then if anything I don't think users would be surprised that the same date remains after switching.

TWLight/users/signals.py
Samwalton9: Ah, yep, you're right. I forgot that we backfilled expiry dates.

jsnshrmn:
There's more test than code here, but I think we're in good shape.

PR #335, opened 2019-10-25
Fix AccessCodes admin view page load timeout (T236413)

uyscuti-wiki:
https://phabricator.wikimedia.org/T236413


Samwalton9:
Looks good to me, though I don't think any fix is too small to be its own PR :)

PR #340, opened 2019-11-01
Re-work partner page metrics

uyscuti-wiki:
[T175564](https://phabricator.wikimedia.org/T175564)


Samwalton9:
It occurs to me that we're also removing CSVs here. That means there's a bunch of CSV URLs we can remove wholesale, and then we just need to extract the partner specific parts of the CSV download views, leaving the rest intact for the metrics page.

jsnshrmn:
I'll try to ask this question in a different way: Why is `get_active_authorizations` using its own logic to get the current number of authorizations in use? Why is it counting some authorizations as active when they are not valid? What is the intended difference between an active authorization and a valid authorization?

uyscuti-wiki:
oh no particular reason. It's just that `get_active_authorizations` was added before `is_valid` and wasn't bothered to update it except for including the date in `date_expires`. Happy to change the method to use `is_valid` if you prefer?

jsnshrmn:
Makes sense. I'm thinking this:
Let's rename `get_active_authorizations` to `get_valid_authorizations`, just so we don't get tripped up on wording differences there.
the `is_valid` property is kind of slow for doing many auths instead of just checking a single auth, so doing a queryset here could be a good performance optimization. Just make sure to use the same logic for getting the set as `is_valid`
https://github.com/WikipediaLibrary/TWLight/blob/master/TWLight/users/models.py#L439
I'd recommend putting in a comment for both the model property and the helper that mentions that the two should be kept in sync.

TWLight/applications/helpers.py
```diff
@@ -170,27 +170,27 @@ def get_output_for_application(app):
 
 
 def get_active_authorizations(partner_pk, stream_pk=None):
```
jsnshrmn: I would have this return the queryset instead of a count. You could add a separate helper for `count_active_authorizations`

TWLight/resources/views.py
```diff
@@ -63,63 +66,27 @@ def get_context_data(self, **kwargs):
...
 
-        context['total_apps'] = Application.objects.filter(
-            partner=partner).count()
-
-        context['total_apps_approved'] = Application.objects.filter(
-            partner=partner, status=Application.APPROVED).count()
-
-        context['total_apps_sent'] = Application.objects.filter(
-            partner=partner, status=Application.SENT).count()
+        all_auths_partner = Authorization.objects.filter(partner=partner)
```
jsnshrmn: assuming you update your get_valid_authorizations in helper to return a valid queryset, you could then just chain the queryset with another filter on partner.pk here
https://docs.djangoproject.com/en/1.11/topics/db/queries/#chaining-filters

TWLight/applications/helpers.py
jsnshrmn: this was basically missed in the initial review of this code: when you have a method or function named get_xxx_objects it should return a list or queryset of those objects

uyscuti-wiki:
@jsnshrmn how does it look now?

TWLight/applications/helpers.py
```diff
@@ -170,38 +171,71 @@ def get_output_for_application(app):
...
+
+
+def count_active_authorizations(partner_pk, stream_pk=None):
+    """
+    Retrieves the numbers of active authorizations using the
+    get_active_authorizations() method above.
+    """
+    if stream_pk:
+        active_authorizations = get_active_authorizations(partner_pk, stream_pk)
+        if active_authorizations is None:
```
jsnshrmn: performance quibble:
This is causing the queryset to be evaluated as an SQL `SELECT`. That means you're deserializing the returned db rows and loading all those objects into memory. You're generating a separate query below when you call `count()`. If you set your variable directly to `get_active_authorizations(partner_pk, stream_pk).count()`, you'll be firing a single `SELECT COUNT` which is much more performant. You can wrap it in a try block if it gets angry when there are no results. Not sure that you need to do that though, as I think it should just return a zero count in that case.

TWLight/applications/helpers.py
```diff
@@ -170,38 +171,71 @@ def get_output_for_application(app):
...
+        # The filter below is equivalent to retrieving all authorizations for a partner
+        # and (or) stream and checking every authorization against the is_valid property
+        # of the authorization model, and hence *must* be kept in sync with the logic in
+        # TWLight.users.model.Authorization.is_valid property.
+        active_authorizations = Authorization.objects.filter(
+            Q(date_expires__isnull=False, date_expires__gte=today) |
+            Q(date_expires__isnull=True),
+            authorizer__isnull=False,
+            user__isnull=False,
+            partner_id__isnull=False,
```
jsnshrmn: `partner=partner_pk` functionally replaces `partner_id__isnull=False` You don't need both. You could put a comment in to that effect.

TWLight/resources/views.py
```diff
@@ -1,9 +1,11 @@
+from datetime import date
+
 from django.contrib import messages
 from django.contrib.auth.models import User
 from django.contrib.auth.decorators import login_required
 from django.core.exceptions import PermissionDenied
 from django.core.urlresolvers import reverse, reverse_lazy
-from django.db.models import Count
+from django.db.models import Count, Q
```
jsnshrmn: unused import?

TWLight/applications/helpers.py
```diff
@@ -170,38 +171,71 @@ def get_output_for_application(app):
...
+            date_authorized__isnull=False,
+            date_authorized__lte=today,
+            partner=partner_pk
+        )
+        if stream_pk:
+            active_authorizations = active_authorizations.filter(stream=stream_pk)
+
+        return active_authorizations
     except Authorization.DoesNotExist:
```
jsnshrmn: Is this an exception that we actually hit? If so return `Authorization.objects.none()` here to return an empty Authorization queryset that way you can always call queryset methods on the return value.

uyscuti-wiki:
Everything seems to be in place. Emphasis on 'seems'.

TWLight/applications/helpers.py
```diff
@@ -169,40 +169,65 @@ def get_output_for_application(app):
...
+            date_authorized__lte=today,
+            partner=partner_pk
+        )
+        if stream_pk:
+            valid_authorizations = valid_authorizations.filter(stream=stream_pk)
+
+        return valid_authorizations
     except Authorization.DoesNotExist:
-        return 0
+        return None
```
jsnshrmn:
```suggestion
        return Authorization.objects.none()
```

uyscuti-wiki:
Thanks Jason for the multiple reviews!

PR #347, opened 2019-11-07
Track application renewal rates

uyscuti-wiki:
[T226495](https://phabricator.wikimedia.org/T226495)

### DIFF
--- a/TWLight/graphs/helpers.py
+++ b/TWLight/graphs/helpers.py
@@ -10,6 +10,8 @@ from django.core.exceptions import FieldError
 from django.utils import timezone
 
 from TWLight.applications.models import Application
+from TWLight.resources.models import Partner
+from TWLight.users.models import Authorization
 
 logger = logging.getLogger(__name__)
 
@@ -274,3 +276,16 @@ def get_users_by_partner_by_month(partner, data_format=JSON):
         return json.dumps(data_series)
     else:
         return data_series
+
+
+def get_proxy_and_renewed_authorizations():
+    proxy_auth = Authorization.objects.filter(partner__authorization_method=Partner.PROXY)
+
+    renewed_auth_ids = []
+    for auth in proxy_auth:
+        latest_app = auth.get_latest_app()
+        if latest_app.parent:
+            renewed_auth_ids.append(auth.id)
+
+    renewed_auth = proxy_auth.filter(id__in=renewed_auth_ids)
+    return proxy_auth, renewed_auth
--- a/TWLight/graphs/helpers.py
+++ b/TWLight/graphs/helpers.py
@@ -49,6 +49,9 @@ def get_median(values_list):
 def get_earliest_creation_date(queryset):
     # Some imported applications had no date and were given
     # creation dates of Jan 1, 1970. This screws up the graphs.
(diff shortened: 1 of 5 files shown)


TWLight/graphs/helpers.py
```diff
@@ -49,9 +50,14 @@ def get_earliest_creation_date(queryset):
     # Some imported applications had no date and were given
     # creation dates of Jan 1, 1970. This screws up the graphs.
     if queryset:
-        earliest_date = queryset.exclude(
-            date_created=datetime.date(1970,1,1)).earliest(
-                'date_created').date_created
+        try:
```
jsnshrmn: Why is this now wrapped in a try block? Comments here would be helpful.

Samwalton9: This feels confusing - we're using this function to operate on both authorization querysets and application querysets, such that we have to rely on `try` to circumvent different field names. Should these be split out into separate functions?

TWLight/graphs/helpers.py
```diff
@@ -80,7 +89,11 @@ def get_data_count_by_month(queryset, data_format=JSON):
             else:
                 timestamp = current_date
 
-            num_objs = queryset.filter(date_created__lte=current_date).count()
+            try:
```
Samwalton9: Likewise here - it's quite opaque reading this why the `try` is necessary. Doing a check for whether we're looking at an `Authorization` queryset or an `Application` one would at a minimum help the readability of this code.

TWLight/graphs/views.py
```diff
@@ -218,7 +234,34 @@ def _write_data(self, response):
         for row in data:
             writer.writerow(row)
 
-            
+class CSVProxyAuthAndRenewals(_CSVDownloadView):
+    def _write_data(self, response):
+        proxy_auth = Authorization.objects.filter(partner__authorization_method=Partner.PROXY)
+        renewed_auth_ids = []
+        for auth in proxy_auth:
```
Samwalton9: We do this a couple of times - can we put it in a helper?

Samwalton9:
I'm not sure we actually need the graph. Having the data as a CSV download is great, but I'm not convinced we learn much from seeing the graph.
+1 to Jason's note about %, could the CSV have an additional column which provides the % (to the closest integer) for ease of access? That's the number we care about for KPIs.
As for the data itself, I don't think the numbers are quite what we're interested in. We have three categories of authorization: Active authorizations, active authorizations based on a renewal, and expired authorizations. We don't care about active authorizations that aren't based on a renewal (they're new ones, not yet requiring renewal). I think the % we actually want is `renewed authorizations/(renewed authorizations + expired authorizations)`.

TWLight/graphs/helpers.py
jsnshrmn: Alternatively, should Authorizations `date_authorized` field be renamed `date_created` so that we're using conventions for this general metadata?

uyscuti-wiki:
Based on [this](https://wikimedia.slack.com/archives/G7EEA63DY/p1573356266000100?thread_ts=1573212926.012800&cid=G7EEA63DY), this should be good to go

PR #348, opened 2019-11-08
Upgrade to Python 3

jsnshrmn:
https://phabricator.wikimedia.org/T209092


jsnshrmn:
I've stripped out a bunch of `utf-8` encode/decode. I recall a that during some of our earlier unicode validation @Samwalton9 manually made changes to introduce non-ascii data into our tests. Any recollection of what that was @Samwalton9? I thought I would add something similar here to make sure we don't introduce a regression, since this upgrade touches 100% of our string handling.

Samwalton9:
> I recall a that during some of our earlier unicode validation @Samwalton9 manually made changes to introduce non-ascii data into our tests. Any recollection of what that was @Samwalton9?

I changed some of the Factory Boy fakers to use a different locale, e.g. for editor usernames

jsnshrmn:
Okay, we're now leveraging faker as a gentle unicode fuzzer. Any time we grab text data from faker, it uses a random locale setting. This includes tests and example data

jsnshrmn:
I ran a restore of production data against this, and it's looking good.

PR #350, opened 2019-11-12
Updated Terms of Use

uyscuti-wiki:
~Once this is live, users need to re-agree to this version of ToU~


uyscuti-wiki:
The `<h3>` tag looks sort of weird with all the white space beneath it. Thought this looked a lot better.

jsnshrmn:
Ah, in that case, I'd recommend adjusting the css for `h3` to get the desired effect. If you are concerned about impacting other `h3` tags on the site, you could check for other usage and see if it improves things, there too. If not, you can always add a class to the `h3` tags on this page and just style those.

uyscuti-wiki:
made a small CSS adjustment to the site-wide headings and at a cursory look at all the pages with said headings look mostly good. In fact this seems to fix [T203765](https://phabricator.wikimedia.org/T203765), which is good. Had to also use some local CSS specific to `terms.html` for the alignment to look more natural and readable.

uyscuti-wiki:
@jsnshrmn the data migration file is in and seems to work as intended. But after running this locally, I wasn't logged our and (or) redirect to the `/terms/ ` page, and I was able to apply for access as well. I'm not sure if the flow would be any different on production.

uyscuti-wiki:
just tested resetting ToU for two of my accounts on production and it seems accounts that have (staff/superusers, not coordinators) might have immunity from agreeing to the ToU. Others would still have to agree to the ToU before they can apply for access.

uyscuti-wiki:
comments updated

jsnshrmn:
> 
> 
> just tested resetting ToU for two of my accounts on production and it seems accounts that have staff/superuser (not coordinator) privileges might have immunity from agreeing to the ToU. Others would still have to agree to the ToU before they can apply for access.
Yep, expected behavior
https://github.com/WikipediaLibrary/TWLight/blob/52384d952c58f4132f9d9c54310f22ff91d8db19/TWLight/view_mixins.py#L258

PR #353, opened 2019-11-17
Fix empty User data on Application Evaluation page

uyscuti-wiki:
[T238489](https://phabricator.wikimedia.org/T238489) 

This PR can be closed if we are to prioritise #348

### DIFF
--- a/TWLight/users/models.py
+++ b/TWLight/users/models.py
@@ -158,8 +158,8 @@ class Editor(models.Model):
     affiliation = models.CharField(max_length=128, blank=True)
 
     def encode_wp_username(self, username):
-        encode_username = username.encode('utf-8')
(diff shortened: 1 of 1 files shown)


TWLight/users/models.py
```diff
@@ -158,8 +158,8 @@ class Meta:
     affiliation = models.CharField(max_length=128, blank=True)
 
     def encode_wp_username(self, username):
-        result = urllib.quote(username)
-        return result
+        _ = username.encode('utf-8')
```
Samwalton9: Just a stylistic thing - `_` is generally used as a variable name when that variable isn't used anywhere, but we do use this below, so it should be named for readability.

uyscuti-wiki:
Fixed by upgrade to Python 3

PR #357, opened 2019-11-21
My Collection page accounts for partner streams and users can choose different streams on renewals

uyscuti-wiki:
[T236456](https://phabricator.wikimedia.org/T236456) and [T230736](https://phabricator.wikimedia.org/T230736)


uyscuti-wiki:
In addition to a review, this needs to thoroughly tested on staging to ensure I didn't miss anything.

Samwalton9:
Just a general note before I take a look through the code that it would be great to separate PRs into one per feature/task - it makes it much easier to review which code belongs to what feature.

uyscuti-wiki:
yeah, this was meant to just focus on the My Collection fix alone, but renewals for streams without the underlying logic seemed rather odd, so T230736 seemed like the natural next step. There shouldn't be much issue reviewing this since everything integrates nicely, but point taken.

uyscuti-wiki:
live on [staging](**url**)

Samwalton9:
I'm not sure that I like the 'Renew' workflow for applying for a different stream. From a UX perspective it's confusing to be clicking Renew in order to apply for something I didn't previously have access to. It's also likely confusing for the reviewer because the user doesn't get a chance to give a new rationale; it copies the rationale from the other application.

Taking a step back, perhaps the best interim solution here is to remove the Apply/Renew/View application logic from the Partner page. It was kind-of a stopgap because we didn't have a Collection page where things like Renewals could be enacted more clearly. I think if we wanted to tell someone to renew their account we'd just send them to their collection page now.

So here's a fresh proposal:
 - Remove the Renew and View Application functionality from partner pages altogether
 - If a partner doesn't have collections and a user has an authorization (active or no) for that partner, grey out the Apply button, with some accompanying explainer text which says something like "View the status of your access in Your Collection"
 - If the partner has collections and the user has authorizations for all of them, do the same grey-out-and-help-text logic as above.
 - If the user doesn't have authorizations for all collections, retain the Apply button
 - If a user applies to a partner with collections and already has authorizations for some (but not all) collections, remove those from the drop-down list of options.

Would that make sense?

uyscuti-wiki:
Updated per the proposal.

uyscuti-wiki:
One caveat with this proposal is it paves the way for duplicate applications.

Samwalton9:
How does it enable duplicate applications?

Also, the message box looks good but I think we should split the logic so the collections message only shows if there are Auths, applications only shows if there are apps (and both if both) :)

uyscuti-wiki:
We don't remove collection options if we don't see corresponding authorizations and we don't disable 'apply' if there are more collections users can apply to. There's no logic in SubmitApplicationView that prevents duplicate applications. Not a deal-breaker, but something to make a note of.

uyscuti-wiki:
Also this might be me forgetting parts of our convo, but I'm not sure how this proposal addresses the "specific title" issue.

Samwalton9:
> We don't remove collection options if we don't see corresponding authorizations

Oh if they apply once, they can apply a 2nd time for the same collection? We should probably check for open apps _or_ auths then.

> Also this might be me forgetting parts of our convo, but I'm not sure how this proposal addresses the "specific title" issue.

Good point, forgot about this aspect. If the partner has `specific_title` the user should continue to be able to apply no matter what, open apps or auths.

Samwalton9:
I was investigating https://phabricator.wikimedia.org/T239314 and realised that we need to update the text in the user renewal notice emails, as it currently points people to the partner pages. We should send users to My Collection instead; https://github.com/WikipediaLibrary/TWLight/blob/master/TWLight/emails/templates/emails/user_renewal_notice-body-text.html

TWLight/applications/forms.py
```diff
@@ -261,8 +265,27 @@ def _add_partner_data_subform(self, partner):
                 if datum == SPECIFIC_STREAM:
                     # Only show streams for this partner
                     partner_id = int(partner[8:])
+                    queryset = Stream.objects.filter(partner_id=partner_id)
```
jsnshrmn: This could use some comments

TWLight/applications/views.py
```diff
@@ -53,6 +53,7 @@
     count_valid_authorizations,
     get_accounts_available,
     is_proxy_and_application_approved,
+    SPECIFIC_STREAM,
```
jsnshrmn: nitpicky, but could you order this so its with the other constants/fields.

jsnshrmn:
hmm:
https://travis-ci.org/WikipediaLibrary/TWLight/builds/645962495#L2507
you might print out the rendered page to your screen to see what's going on here.

uyscuti-wiki:
Turns out, there's no zero-padding before the date, which managed to skip my eye since all the dates I previously ran the tests were two digits.

PR #363, opened 2019-11-25
Always allow users to extend lifetime of an authorization (T238671)

Samwalton9:
Shuffles the logic a bit so we never hide the Renew button, instead always displaying it with the (in my opinion) slightly clearer text 'Extend'. If the authorization expires, we switch back to 'Renew'.

### DIFF
--- a/TWLight/users/models.py
+++ b/TWLight/users/models.py
@@ -422,7 +422,7 @@ class Authorization(models.Model):
         app_label = "users"
         verbose_name = "authorization"
         verbose_name_plural = "authorizations"
-        unique_together = ("user", "partner", "stream", "date_authorized")
+        unique_together = ("user", "partner", "stream")
 
     coordinators = get_coordinators()
 
--- a/TWLight/applications/models.py
+++ b/TWLight/applications/models.py
@@ -333,6 +333,23 @@ class Application(models.Model):
                 )
         return user_instructions
 
+    def get_authorization(self):
+        """
+        For a given application, find an authorization for this partner-stream-editor, if possible.
+        """
+        authorization = Authorization.objects.filter(
+                partner=self.partner, user=self.editor.user
+            )
+        if self.specific_stream:
+            authorization = authorization.filter(
+                stream=self.specific_stream
+            )
+
+        if authorization.exists():
+            return authorization[0]
+        else:
+            return None
+
     def is_instantly_finalized(self):
         """
         Check if this application is to a partner or collection for which
--- a/TWLight/applications/models.py
+++ b/TWLight/applications/models.py
@@ -337,18 +337,19 @@ class Application(models.Model):
         """
         For a given application, find an authorization for this partner-stream-editor, if possible.
         """
-        authorization = Authorization.objects.filter(
-                partner=self.partner, user=self.editor.user
-            )
-        if self.specific_stream:
-            authorization = authorization.filter(
-                stream=self.specific_stream
-            )
-
-        if authorization.exists():
-            return authorization[0]
-        else:
+        try:
+            if self.specific_stream:
+                authorization = Authorization.objects.get(
+                    partner=self.partner, user=self.editor.user, stream=self.specific_stream
+                )
+            else:
+                authorization = Authorization.objects.get(
+                    partner=self.partner, user=self.editor.user
+                )
+        except Authorization.DoesNotExist:
             return None
+
+        return authorization
 
     def is_instantly_finalized(self):
         """
--- a/TWLight/applications/templates/applications/application_evaluation.html
+++ b/TWLight/applications/templates/applications/application_evaluation.html
@@ -194,6 +194,17 @@
         {% trans 'No' %}
       {% endif %}
     </div>
+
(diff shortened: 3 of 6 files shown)


Samwalton9:
A note adjacent to this PR - if a proxy list is full can we renew someone's access? Or do we stop ourselves from doing so?

uyscuti-wiki:
Couple of things: 

1. When we say "extend", I think users would expect an extension on their access, but currently the access is [reset](https://github.com/WikipediaLibrary/TWLight/blob/master/TWLight/applications/models.py#L492) to the requested duration i.e. if a user requests an extension for 6 months of their access, which has 6 more months left, the request effectively is void since nothing changes. 

2. Since we are allowing extensions, coordinators probably should be able to see the status of the applicants current access (time remaining), and change the duration if necessary after consulting with the user, to allow them to make more informed decisions. This is particularly true when there's a scarcity of resources – prioritise applications with no access over applications with some time left.



> A note adjacent to this PR - if a proxy list is full can we renew someone's access? Or do we stop ourselves from doing so?

We [already do this](https://github.com/WikipediaLibrary/TWLight/blob/master/TWLight/applications/views.py#L815). Partners are marked 'Waitlisted' and approvals are disallowed.

Samwalton9:
>When we say "extend", I think users would expect an extension on their access, but currently the access is reset to the requested duration i.e. if a user requests an extension for 6 months of their access, which has 6 more months left, the request effectively is void since nothing changes.

Good point! I want to see if we can solve this with a wording change rather than a functionality change - the idea is that users should be able to refresh the duration of their access up to 12 months from now rather than being able to keep adding time indefinitely. I agree that 'Extend' indicates that the time is additive. We want something that implies editing the expiry date rather than adding to it.

> Since we are allowing extensions, coordinators probably should be able to see the status of the applicants' current access (time remaining), and change the duration if necessary after consulting with the user, to allow them to make more informed decisions. This is particularly true when there's a scarcity of resources – prioritise applications with no access over applications with some time left.

This is a nice idea. Probably best to just include a row on the evaluate application page for 'current expiry date' or somesuch on renewals.

>if there's a shortage, there should be a fair chance for everyone. By which I mean renewals go to the end of the list and fresh applications get priority.

I think this is something we want to discuss. Personally I'd argue that users who currently have access should be given priority - the chance a user who is requesting renewal will use their access is higher than a fresh applicant I think.

Samwalton9:
>I want to see if we can solve this with a wording change rather than a functionality change

Just checked and when users click Extend the wording on the renewal page is "The number of months you wish to have this access for before renewal is required". I feel like this might be fine, that seems clear to me.

>Probably best to just include a row on the evaluate application page for 'current expiry date' or somesuch on renewals.

Added a commit for this, in addition to a new `get_authorization` method which finds an authorization for an application, i.e. the reverse of `get_application` on authorizations. I also realised `get_application` isn't considering streams, so we should fix that.

uyscuti-wiki:
For a consistent UX, I'd suggest relabelling the renewal form field for access duration and doing some calculation to pop out the options that if added would extend the time of an auth by more than a year with a note stating a user cannot extend an auth for more than a year. And if the auth cannot, at a minimum, add one more month to its expiry, return the user back to 'My Collection' with a warning message.

Samwalton9:
> I'd suggest relabelling the renewal form field for access duration and doing some calculation to pop out the options that if added would extend the time of an auth by more than a year with a note stating a user cannot extend an auth for more than a year

This would require us to change the behaviour of the renewal form - from updating the expiry to X months from now to _adding_ X months to the expiry. I think the current workflow is clear.

Samwalton9:
In `get_authorization` we should probably just use `.get` instead of `.filter`, but I should double check the data structure to see at what level we 'know' there's only one partner-user-stream combination.

Samwalton9:
> I should double check the data structure to see at what level we 'know' there's only one partner-user-stream combination

We do `unique_together = ("user", "partner", "stream", "date_authorized")`, which isn't quite the same - we relaxed the constraint somewhat to include `date_authorized`. This is unnecessary because there shouldn't be multiple auths on different days either. As such, I've taken that out in the following commit, while adjusting `filter` to `get` since we know that combination is unique.

TWLight/users/models.py
```diff
@@ -422,7 +422,7 @@ class Meta:
         app_label = "users"
         verbose_name = "authorization"
         verbose_name_plural = "authorizations"
-        unique_together = ("user", "partner", "stream", "date_authorized")
```
jsnshrmn: Yep, this was one of the earliest bits of logic for authorizations, and our use-case for them has dramatically firmed up since then.

Samwalton9:
Since we didn't really change any functionality here, this should be covered by existing tests. Should. I'll write some tests.

Samwalton9:
The act of extending a proxy authorization is already tested in https://github.com/WikipediaLibrary/TWLight/blob/4ae6217da3d9de5dd58dd46a056def524a6a2763/TWLight/applications/tests.py#L1992, though it didn't explicitly check that the expiry date was set correctly. I'll amend that and add a test for `get_authorization`.

PR #356, opened 2019-11-20
Change ToU UX and ensure users don't access proxied resources without first agreeing to the ToU

uyscuti-wiki:
[T238656](https://phabricator.wikimedia.org/T238656)

PR #368, opened 2019-12-03
Fix notification pending applicants tou changes

jsnshrmn:


uyscuti-wiki:
I'm not sure how to approach this in a more generalised way, but ideally we don't want to leave comments on the app that already have comments in them or the coordinator has left a comment, and the comments were made later than when #356 was merged.

jsnshrmn:
That's great feedback. I updated the command so that apps in which 'TWL Team' has already left a comment are excluded.

TWLight/applications/management/commands/notify_applicants_tou_changes.py
```diff
@@ -20,20 +23,28 @@ def handle(self, *args, **options):
...
-                comment=_("Our terms of use have changed. "
-                          "Your applications will not be processed until you log in and agree to our updated terms.")
-            )
-            comment.save()
+            if (
+                Comment.objects.filter(
+                    object_pk=str(app.pk), site_id=settings.SITE_ID, user=twl_team
+                ).count()
+                == 0
+            ):
```
uyscuti-wiki: We have cases [like this](https://wikipedialibrary.wmflabs.org/applications/evaluate/10573/) where the coordinator responsible already has left a comment. Is there a reason we don't want to check for coordinator comments?

jsnshrmn: Because there are lots of reasons for coordinators to comment on applications besides our terms of use changing. We will be changing our terms again in the future, so this should be a generally working tool.

jsnshrmn: I could add a check for the last commented posted date. eg. maybe we only comment if the app meets the existing criteria and has also not had a comment in n days?

uyscuti-wiki: yes, that would be awesome! We give coordinators the applicant's ToU info, which they should consider before approving an app, so this leaves some room for coordinator discretion.

uyscuti-wiki: Not really a discretion if we think about it, more of a trusting coordinators to take care of it.

jsnshrmn: Okay, added in a check to see if there were comments in in the last 8 days. No comment will be left if TWL has ever commented OR if anyone has commented in the last 8 days. If I run this command tomorrow, it should not comment on any applications that have had comments since the command was run the first time.

PR #370, opened 2019-12-05
Add notice on AppEval when there's a possibility of an app being waitlisted 

uyscuti-wiki:
[T162309](https://phabricator.wikimedia.org/T162309)


uyscuti-wiki:
This, I believe, is ready for another pass.

TWLight/applications/views.py
```diff
@@ -989,11 +990,23 @@ def get_form(self, form_class=None):
...
             form.fields["status"].choices = status_choices
 
+        if app.editor.user == self.request.user:
+            if more_applications_than_accounts_available(app):
+                messages.add_message(
+                    self.request,
+                    messages.WARNING,
+                    # fmt: off
+                    # Translators: This warning is message is shown to applicants when the number of pending applications is greater than the number of accounts available.
+                    _("There are more number of pending applications than there are accounts available. Your application might get waitlisted."),
```
jsnshrmn, line 1006: wording suggestion
```suggestion
                    _("There are more pending applications than available accounts. Your application might get waitlisted."),
```

TWLight/applications/tests.py
```diff
@@ -3409,6 +3409,27 @@ def test_no_status_form_for_bundle_partners(self):
...
+            self.user.save()
+            self.user.userprofile.terms_of_use = True
+            self.user.userprofile.save()
+
+            self.client.login(username=self.user.username, password="editor")
+            self.message_patcher.stop()
+            response = self.client.post(self.url)
+            messages = list(response.context["messages"])
+            self.assertEqual(len(messages), 1)
+            self.assertEqual(
```
jsnshrmn, line 3426: This test is pretty brittle since it depends on the entirety of the message staying the same. Instead of testing the final output of the view, you could just test the true/false response of the helper function by feeding it a few apps and testing for true/false output.

uyscuti-wiki: Makes sense. Fixed!

jsnshrmn:
Looks great. Thanks!

PR #378, opened 2019-12-12
Allow coordinators to see unavilable partners they are assigned to

HAKSOAT:
The current code only lets staff to see unavailable partners they are assigned to.
The coordinators may have reasons to see unavailable partners. One of these reasons could be for proofreading.

Bug: T215273

### DIFF
--- a/TWLight/resources/tests.py
+++ b/TWLight/resources/tests.py
@@ -221,6 +221,30 @@ class PartnerModelTests(TestCase):
 
         self.assertContains(response, partner.get_absolute_url())
 
+    def test_visibility_of_not_available_5(self):
+        """
+        Coordinators *should* see NOT_AVAILABLE partner pages.
+        We won't test to see if they include the message, because RequestFactory
+        doesn't include message middleware, but we can't use the Django test
+        client to log in users without passwords (which is our normal user).
+        """
+        partner = PartnerFactory(status=Partner.NOT_AVAILABLE)
+        detail_url = partner.get_absolute_url()
+
+        editor = EditorFactory()
+        coordinators = get_coordinators()
+        coordinators.user_set.add(editor.user)
+
+        partner.coordinator = editor.user
+        partner.save()
+
+        request = RequestFactory().get(detail_url)
+        request.user = editor.user
+
+        # This should not raise Http404.
+        response = PartnersDetailView.as_view()(request, pk=partner.pk)
+        self.assertEqual(response.status_code, 200)
+
     def test_review_app_page_excludes_not_available(self):
         partner = PartnerFactory(status=Partner.NOT_AVAILABLE)
         _ = ApplicationFactory(partner=partner, status=Application.PENDING)
--- a/TWLight/resources/views.py
+++ b/TWLight/resources/views.py
@@ -51,13 +51,8 @@ class PartnersDetailView(DetailView):
 
     def get_context_data(self, **kwargs):
         context = super(PartnersDetailView, self).get_context_data(**kwargs)
-
         partner = self.get_object()
-
         if partner.status == Partner.NOT_AVAILABLE:
-            # This should be guaranteed by get_queryset and the manager
-            # definitions.
-            assert self.request.user.is_staff
             # Translators: Staff members can view partner pages which are hidden from other users. This message appears on those specific partner resource pages.
             messages.add_message(
                 self.request,
@@ -145,10 +140,27 @@ class PartnersDetailView(DetailView):
         return context
 
     def get_queryset(self):
+        # We have three types of users who might try to access partner pages - the partner's coordinator, staff,
+        # and normal users. We want to limit the list of viewable partner pages in different ways for each.
+
+        # By default users can only view available partners
+        available_partners = Partner.objects.all()
+        partner_list = available_partners
+
+        # If logged in, what's the list of unavailable partners, if any, for which the current user is the coordinator?
+        if self.request.user.is_authenticated():
+            coordinator_partners = Partner.even_not_available.filter(
+                coordinator=self.request.user, status=Partner.NOT_AVAILABLE
+            )
+            if coordinator_partners.exists():
+                # Coordinated partners are also valid for this user to view
+                partner_list = available_partners | coordinator_partners
+
         if self.request.user.is_staff:
-            return Partner.even_not_available.order_by("company_name")
-        else:
-            return Partner.objects.order_by("company_name")
+            # Staff can see any partner pages, even unavailable ones.
+            partner_list = Partner.even_not_available.all()
+
+        return partner_list
 
 
 class PartnersToggleWaitlistView(CoordinatorsOnly, View):


Samwalton9:
Looks good to me. Ideally we should also have a test for this functionality. Would you be interested in writing a test @HAKSOAT?

HAKSOAT:
Yes, I'll work on it.

HAKSOAT:
Hello @Samwalton9 I'm finding it hard to figure out how to go about writing the test. I'll appreciate some help.

Samwalton9:
You can view a test for staff being able to see a not available partner at https://github.com/WikipediaLibrary/TWLight/blob/master/TWLight/resources/tests.py#L186. If you copied that to create a new function below (renaming it to something like `test_visibility_of_not_available_5`), you could do something like the following:

Remove:
```
editor = EditorFactory()
editor.user.is_staff = True
editor.user.save()
```

Instead, assign the user as a coordinator to the partner:
```
editor = EditorFactory()
coordinators = get_coordinators()
coordinators.user_set.add(editor.user)

partner.coordinator = editor.user
partner.save()
```

I haven't tested this, so you might need to run the tests and check that this works :)

HAKSOAT:
Thanks a lot. That's almost all of the testing done.

Samwalton9:
This test looks great, and it's uncovered a bug! It looks like we also need to change this line:

https://github.com/WikipediaLibrary/TWLight/blob/master/TWLight/resources/views.py#L148

to `self.request.user.is_staff or self.request.user == partner.coordinator` in the same way.

HAKSOAT:
Phew! Thanks for coming to my aid. The proposed change causes more tests to fail though.

I add to do `partner=self.get_object()` in the `get_queryset()` method before changing that line. I'm not sure if I did this the right way.

Samwalton9:
Ah you're totally right - we hadn't defined `partner` so `partner=self.get_object()` would make sense, but I also see how that causes a circular dependency!

So this might require a more involved solution. One thing that occurs to me is that we appear to be validating that the partner is available twice - once by limiting the queryset of available objects, and once by doing `if partner.status == Partner.NOT_AVAILABLE:`. @jsnshrmn - would you see an issue with removing the `get_queryset()` logic (in favour of having it always return `even_not_available` partners, and simply relying on this check:https://github.com/WikipediaLibrary/TWLight/blob/master/TWLight/resources/views.py#L57 ?

HAKSOAT:
I don't fully understand what the question posed to @jsnshrmn implies. However, I checked the models file and saw the `even_not_available` field. So I guess I'll just wait for a reply from Jason.

HAKSOAT:
Hello @Samwalton9 Please I need some help here.

jsnshrmn:
@HAKSOAT sorry for the delay here, I've been out of touch for a few weeks. Basically, @Samwalton9's concerns relate to security and maintainability. We use a partner coordinator check in several places, sometimes for ui purposes (only show the user what they need) and sometimes for data security (don't show them data they are not privy to). I'll have a look and offer some feedback on this sometime this week.

HAKSOAT:
Thanks @jsnshrmn I could remind you by Friday. I'll pick up a new task in the meantime.

jsnshrmn:
Yeah, my previous review was based on a read through only, I'd have to play with this to see why the `get_queryset` method is getting called at all, but I wouldn't surprised if it's a result of us swapping out the model manager. I believe @Samwalton9 is suggesting one of two things:

either this:
```
    def get_queryset(self):
        if self.request.user.is_staff or self.request.user == partner.coordinator
            return Partner.even_not_available.order_by("company_name")
        else:
            return Partner.objects.order_by("company_name")
```
or this
```
    def get_queryset(self):
        return Partner.even_not_available.order_by("company_name")
```

The first option would extend the current approach, which looks to be a bit of belt and suspenders. The second option (If I'm correct) might still get us the same end result with simpler code. Both of these are untested, so you should play with them.

HAKSOAT:
Thanks @jsnshrmn I'll give feedback soon.

jsnshrmn:
Now that I'm looking through this more carefully, I can see that we do in fact override `get_queryset` for all the class-based views where we've changed the model manager. I think my preference actually be to have that guard clause to be in `get_queryset` and not in `get_context_data` if we don't need both.

Samwalton9:
Agreed with Jason that the `get_queryset` modification seems the most sensible. We even wrote a comment that confirms the other code is effectively redundant! So, all that said, here seems to be the best way forward @HAKSOAT:

Remove this `assert` line and the comment right above it: https://github.com/WikipediaLibrary/TWLight/blob/master/TWLight/resources/views.py#L60

Then change the `get_queryset` function to this, per Jason's comment:
```
    def get_queryset(self):
        if self.request.user.is_staff or self.request.user == partner.coordinator
            return Partner.even_not_available.order_by("company_name")
        else:
            return Partner.objects.order_by("company_name")
```

This should solve the issue :)

Samwalton9:
Ah, whoops, we also need `partner = self.get_object()` in there, since we haven't defined `partner` yet!

HAKSOAT:
> 
> 
> Ah, whoops, we also need `partner = self.get_object()` in there, since we haven't defined `partner` yet!

Yes. I realized. I went with the second solution already.

HAKSOAT:
Ooops. Failed too.

HAKSOAT:
Failed too. I think the issue is that the `get_object` calls `get_queryset`. I really need to get the environment working properly locally, so I can stop testing with every push.

Samwalton9:
Oh, right, my bad - we already discussed that above! We'll discuss this and see if we can figure out the best solution, this turned out to be a frustrating little task 😁 

In the meantime, I agree getting a development environment up and running would be a great idea. Have you attempted to do so yet? I'm more than happy to help with any issues you're running into there.

Samwalton9:
Alright I spent a little time playing around with this today and came up with a solution. We want to handle each case almost individually in `get_queryset()` without specifying any particular partner. This worked in my local testing:
```
    def get_queryset(self):
        # We have three types of users who might try to access partner pages - the partner's coordinator, staff,
        # and normal users. We want to limit the list of viewable partner pages in different ways for each.

        # By default users can only view available partners
        available_partners = Partner.objects.all()
        partner_list = available_partners

        # If logged in, what's the list of unavailable partners, if any, for which the current user is the coordinator?
        if self.request.user.is_authenticated():
            coordinator_partners = Partner.even_not_available.filter(
                coordinator=self.request.user, status=Partner.NOT_AVAILABLE
            )
            if coordinator_partners.exists():
                # Coordinated partners are also valid for this user to view
                partner_list = available_partners | coordinator_partners

        if self.request.user.is_staff:
            # Staff can see any partner pages, even unavailable ones.
            partner_list = Partner.even_not_available.all()

        return partner_list
```

HAKSOAT:
Hello @Samwalton9 Thanks for the help. I was not available for a while. About my setup issues, I get the following:

Screenshot (66)

Screenshot (67)

HAKSOAT:
I cloned the repo again and gave it another shot. Got thi:

Creating twlight_db_1 ... error                                                                                         
ERROR: for twlight_db_1  Cannot create container for service db: invalid volume specification: 'C:\Users\HAKS\DevStuff\WikiMedia\TWLight\secrets\DJANGO_DB_NAME:/run/secrets/DJANGO_DB_NAME:ro'

ERROR: for db  Cannot create container for service db: invalid volume specification: 'C:\Users\HAKS\DevStuff\WikiMedia\TWLight\secrets\DJANGO_DB_NAME:/run/secrets/DJANGO_DB_NAME:ro'
ERROR: Encountered errors while bringing up the project.

Samwalton9:
Hm, this is the same error @NicholusMuwonge had at https://phabricator.wikimedia.org/T218857. I'm not sure if this ever got solved.

There's a potential solution at https://www.computerhope.com/issues/ch000549.htm, and more information and suggestions at https://github.com/docker/compose/issues/5842. Give those a go and see if you can get them to work?

Either way, I think the code snippet I pasted above should work for you :)

jsnshrmn:
Hmm @HAKSOAT are you using docker desktop or docker toolbox? So far all the messiest problems we've run into have been with users on docker toolbox, which I don't think is supported anymore.

HAKSOAT:
I'm using the docker toolbox

HAKSOAT:
I'll take a look at the links @Samwalton9 shared and see if I can fix the problem.

HAKSOAT:
I think I just found a solution. I created a Ubuntu VM and installed docker and docker compose there. What command do I use to run the tests? @Samwalton9

Samwalton9:
Oh great! Run `docker-compose exec twlight /app/bin/virtualenv_test.sh`

HAKSOAT:
@Samwalton9 Finally it passed. Feels like I did nothing though. Hehe. But a good way to start. Thanks for the guidance and assistance.

Samwalton9:
Hooray! Haha, don't worry - this turned out to be more complicated than I anticipated, and this is your first contribution to the repository, we don't expect you to know how everything works right away.

Thanks for your contribution :)

PR #381, opened 2019-12-16
Allow coordinators to customise which types of applications count towards email reminders

uyscuti-wiki:
[T186004](https://phabricator.wikimedia.org/T186004)


uyscuti-wiki:
Just a word of caution: the changes in this PR gave me some headache with migrations locally since I've made changes to the UserProfile model. The error was about a reference made from the automatically created migration `auth.xxxx_proxyauthorization` to the migration `users.xxxx_some_migration`, that the reference was being made before there was even a  `users.xxxx_some_migration`. This is live on [staging](http://twlight-staging.wmflabs.org/users/) where it seemed to have worked fine and this could very well be an isolated incident due to how I treated the migration files. But a heads up in any case so that we don't end up breaking production.

jsnshrmn:
I should say that the template markup is adequately commented as well.

PR #382, opened 2019-12-17
Back to partner button added

soumyaa1804:
Back to partner button added and some indentations changed.


Samwalton9:
Thanks @soumyaa1804! Could you include a screenshot of what this looks like? It would save us time in pulling the changes and checking them locally :)

TWLight/resources/templates/resources/partner_detail.html
```diff
@@ -22,23 +22,24 @@
...
+        {% comment %} Translators: This text labels partners who have no available accounts remaining and therefore have a waitlist for access {% endcomment %}
+        <span class="resource-label">{% trans "Waitlisted" %}</span>
+        <p>
+        {% comment %} Translators: If we have no available accounts for a partner, the coordinator can change the application system to a waiting list. {% endcomment %}
+        {% blocktrans trimmed %}
+        There are no access grants available for this partner
+        at this time. You can still apply for access; applications will
+        be processed when access is available.
+        {% endblocktrans %}
+        </p>
```
uyscuti-wiki: We stick to two spaces for indentation in all the templates. Would be nice if you could revert the addition of one more space from line 33 through 42.

TWLight/resources/templates/resources/partner_detail.html
```diff
@@ -22,23 +22,24 @@
 {% endblock extra_css %}
 
 {% block content %}
+  <a href="{% url 'partner:filter' %}" class="btn btn-primary">Back to partners</a>
```
uyscuti-wiki: This could be a `btn-default` rather than `btn-primary` and could be moved beneath `<h1>{{ object }}</h1>` so as to blend even more nicely.

TWLight/resources/templates/resources/partner_detail.html
soumyaa1804: Sure! would do that in a minute.

TWLight/resources/templates/resources/partner_detail.html
soumyaa1804: Okay!

soumyaa1804:
> Thanks @soumyaa1804! Could you include a screenshot of what this looks like? It would save us time in pulling the changes and checking them locally :)

I tried to check the changes locally but on browsing through any partner, it is giving this error. 

image

uyscuti-wiki:
Should look like this 
image

@soumyaa1804 that's weird. Did you not use OAuth to login to your local setup?

soumyaa1804:
> Should look like this
> image
> 
> @soumyaa1804 that's weird. Did you not use OAuth to login to your local setup?

I suppose I did.

Screenshot from 2019-12-17 20-09-06

Samwalton9:
That server error might be related to https://phabricator.wikimedia.org/T226648, i.e. when there are no coordinators, but I could be wrong.

Anyway, it's evidently not the fault of your changes so we'll ignore it for now!

Your changes look good, though we need to add translation support for that new text. That can be done with the trans tag, like `{% trans 'text' %}`. We also need a comment on the line above providing some context for where the text is located. In this case it would be something like "Translators: This text is located on individual partner pages, and when clicked takes users back to the list of publishers."
You can see this comment and trans tag behaviour at https://github.com/WikipediaLibrary/TWLight/blob/master/TWLight/templates/dashboard.html#L17 :)

soumyaa1804:
> That server error might be related to https://phabricator.wikimedia.org/T226648, i.e. when there are no coordinators, but I could be wrong. Anyway, it's evidently not the fault of your changes so we'll ignore it for now!
> 
> Your changes look good, though we need to add translation support for that new text. That can be done with the trans tag, like `{% trans 'text' %}`. We also need a comment on the line above providing some context for where the text is located. In this case it would be something like "Translators: This text is located on individual partner pages, and when clicked takes users back to the list of publishers."
> You can see this comment and trans tag behaviour at https://github.com/WikipediaLibrary/TWLight/blob/master/TWLight/templates/dashboard.html#L17 :)

I just cloned it again and it worked!
image

image

Samwalton9:
Oh great. Looks good to me, thanks for your contribution!

PR #386, opened 2020-01-07
When finding the latest application to renew, ignore denied applications

uyscuti-wiki:
[T239505](https://phabricator.wikimedia.org/T239505)

### DIFF
--- a/TWLight/applications/models.py
+++ b/TWLight/applications/models.py
@@ -452,6 +452,16 @@ def post_revision_commit(sender, instance, **kwargs):
         instance.status = Application.SENT
         instance.save()
 
+    # Renewals are for applications that are approved/sent.
+    # Having a parent for NOT_APPROVED apps hinders us from
+    # correctly renewing the parent. So, we unset the parent
+    # if the status is NOT_APPROVED and the app already has
+    # a parent.
+    if instance.status == Application.NOT_APPROVED:
+        if instance.parent:
+            instance.parent = None
+            instance.save()
+
     # Authorize editor to access resource after an application is saved as sent.
 
     if instance.status == Application.SENT:
--- a/TWLight/applications/models.py
+++ b/TWLight/applications/models.py
@@ -457,10 +457,9 @@ def post_revision_commit(sender, instance, **kwargs):
     # correctly renewing the parent. So, we unset the parent
     # if the status is NOT_APPROVED and the app already has
     # a parent.
-    if instance.status == Application.NOT_APPROVED:
-        if instance.parent:
-            instance.parent = None
-            instance.save()
+    if instance.status == Application.NOT_APPROVED and instance.parent:
+        instance.parent = None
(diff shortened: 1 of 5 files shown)


Samwalton9:
Could it make more sense to update or copy `get_latest_app()` and instead have a `get_latest_sent_app()` function, rather than having all that logic in the view?

uyscuti-wiki:
don't think so; we need the latest app, based on which we make other calculations.

Samwalton9:
My initial thought was, if we only use `get_latest_app()` for finding the latest sent application, we could just modify it for clarity. The other option being to make a brand new `get_latest_sent_app()` that's exactly the same as get_latest_app with the additional status filter.

uyscuti-wiki:
@Samwalton9 I've factored out the logic like you've requested, and added in something that somehow missed my eye the first time. Can you take a look?

TWLight/applications/models.py
```diff
@@ -452,6 +452,15 @@ def post_revision_commit(sender, instance, **kwargs):
...
         instance.save()
 
+    # Renewals are for applications that are approved/sent.
+    # Having a parent for NOT_APPROVED apps hinders us from
+    # correctly renewing the parent. So, we unset the parent
+    # if the status is NOT_APPROVED and the app already has
+    # a parent.
+    if instance.status == Application.NOT_APPROVED and instance.parent:
+        instance.parent = None
+        instance.save()
```
Samwalton9: Oh that's clever. This hadn't occurred to me :)

TWLight/users/views.py
```diff
@@ -690,9 +688,15 @@ def get_context_data(self, **kwargs):
...
-                        if not each_authorization.latest_app.is_renewable:
+                    latest_app = each_authorization.get_latest_app()
+                    if latest_app:
+                        if latest_app.status != Application.SENT:
+                            each_authorization.latest_sent_app = (
+                                each_authorization.get_latest_sent_app()
+                            )
+                        else:
+                            each_authorization.latest_sent_app = latest_app
+                        if not latest_app.is_renewable:
```
Samwalton9: Do we need to do all this? Why can't we just use `get_latest_sent_app()` right away?

uyscuti-wiki: I did think about this and [apparently](https://github.com/WikipediaLibrary/TWLight/blob/master/TWLight/applications/models.py#L390) APPROVED applications can be renewed too. So, didn't want to break anything, and open to suggestions.

uyscuti-wiki: Also, we cannot stop if we don't find any SENT apps. We'd want to know to know if there are other apps that fit our criteria for which we'd need `get_latest_app()` here.

Samwalton9: Well that's weird. I don't know why we would want to let folks renew an approved application. Will ping Nikki on Slack, we might want to adjust that.

PR #387, opened 2020-01-08
Update terms.html

snikithasiddavatam:


Samwalton9:
Thanks for this pull request! It looks like you've successfully changed the accompanying help text for this, but haven't changed the text on the button itself. That text can be found at https://github.com/WikipediaLibrary/TWLight/blob/master/TWLight/users/forms.py#L129 :)

TWLight/users/forms.py
```diff
@@ -126,7 +126,7 @@ def __init__(self, *args, **kwargs):
         self.helper.layout = Layout(
             "terms_of_use",
             # Translators: this 'I accept' is referenced in the terms of use and should be translated the same way both places.
-            Submit("submit", _("I accept"), css_class="btn btn-default"),
+            Submit("submit", _("submit"), css_class="btn btn-default"),
```
Samwalton9: Could you capitalise this? So `Submit` instead of `submit`. Thanks :)

TWLight/users/forms.py
```diff
@@ -30,7 +30,7 @@ def __init__(self, *args, **kwargs):
         self.helper = FormHelper()
         self.helper.add_input(
             Submit(
-                "submit",
+                "Submit",
```
uyscuti-wiki: Don't think this does what you probably think it does. The first argument in `layout.Submit()` is the `name` of the submit button, which is used to reference the correct button on POST and is not user facing. That would be the next argument (`"Update profile"` in this case), which forms the `value` of the button, which _is_ user facing. And by convention, names, ids, classes etc are typically in lowercase, to improve readability and to avoid confusion.

TWLight/users/forms.py
```diff
@@ -126,7 +126,7 @@ def __init__(self, *args, **kwargs):
         self.helper.layout = Layout(
             "terms_of_use",
             # Translators: this 'I accept' is referenced in the terms of use and should be translated the same way both places.
-            Submit("submit", _("I accept"), css_class="btn btn-default"),
+            Submit("Submit", _("Submit"), css_class="btn btn-default"),
```
uyscuti-wiki: Ditto. Someone unacquainted with the codebase/crispy forms will have trouble understanding which one of the two submits is the name and which one is the value. Hence it's better to have the `name` (first argument) in all lowercase.

TWLight/users/forms.py
```diff
@@ -155,6 +155,6 @@ def __init__(self, user, *args, **kwargs):
             "use_wp_email",
             # Translators: This labels a button which users click to change their email.
             Submit(
-                "submit", _("Update email"), css_class="btn btn-default col-md-offset-2"
+                "Submit", _("Update email"), css_class="btn btn-default col-md-offset-2"
```
uyscuti-wiki: Ditto.

Samwalton9:
@snikithasiddavatam It looks like you've just got a couple of things to fix in this PR, did you need any further clarification or feedback?

Samwalton9:
Closing this ticket based on lack of activity for requested changes.

PR #388, opened 2020-01-09
Update forms.py

snikithasiddavatam:


Samwalton9:
Can you commit this change to the existing pull request (https://github.com/WikipediaLibrary/TWLight/pull/387), rather than a new one? Thanks :)

PR #393, opened 2020-01-17
Tidy up unused code and imports, update comments

Samwalton9:
Just did a general pass through the code to ensure comments were (relatively) up to date and tidying up some unused or redundant imports and code.

There shouldn't be anything particularly controversial in here but would appreciate a once-over to verify.


Samwalton9:
Merging now that this PR is only changes to comments, unused imports, and unused files.

PR #396, opened 2020-01-22
Fix requested_access_duration check and knock-on test failures

Samwalton9:
When generating authorizations we were checking for `requested_access_duration is None` on the submitted application. Users can't submit applications with `None` data for that field on a proxy partner unless the partner has `requested_access_duration is False`*, so we should check for that directly.

I made that change and then discovered this had a knock-on effect of causing failures on a bunch of tests because they weren't setting the application `requested_access_duration` field on app submission (they were sending `None`), so I went through and updated those applications to send the correct data.

*We should name those fields differently, it's very confusing.

### DIFF
--- a/TWLight/applications/factories.py
+++ b/TWLight/applications/factories.py
@@ -3,6 +3,7 @@
 import factory
 
 from TWLight.resources.factories import PartnerFactory
+from TWLight.resources.models import Partner
 from TWLight.users.factories import EditorFactory
 
 from TWLight.applications.models import Application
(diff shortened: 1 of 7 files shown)


Samwalton9:
Not a priority PR, this is primarily a test-impacting change.

Samwalton9:
As part of this I should just go ahead here and rename `requested_access_duration` to `needs_access_duration` on the partner model - it was super confusing that these fields were named the same.

Samwalton9:
Nevermind, turns out that's by design... We should fix that.

Samwalton9:
This shouldn't be merged. The original case is actually desirable for the transition period for authentication-based access.

We should review what we actually want to fix here another time.

PR #400, opened 2020-02-06
daily cron task for updating editor eligibility as needed.

jsnshrmn:
Note that this a PR against staging, since that's where we're doing bundle-related work.


jsnshrmn:
Merging, but not deleting branch

PR #405, opened 2020-02-14
Show Error on 'Mark as Sent' page for automatic-send partners

lalit97:
If a Partner's authorization method is Link, Proxy or Bundle
then the applications are sent automatically. So the 'Mark as Sent'
page is no longer needed for that Partner.

Bug: T234551

### DIFF
--- a/TWLight/applications/views.py
+++ b/TWLight/applications/views.py
@@ -1169,7 +1169,7 @@ class SendReadyApplicationsView(PartnerCoordinatorOnly, DetailView):
                 request, *args, **kwargs
             )
         else:
-            raise Http404("Application's for this Partner are sent automatically")
+            raise Http404("Applications for this Partner are sent automatically")
 
     def get_context_data(self, **kwargs):
         context = super(SendReadyApplicationsView, self).get_context_data(**kwargs)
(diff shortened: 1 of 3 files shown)


TWLight/applications/views.py
```diff
@@ -1161,6 +1161,16 @@ class SendReadyApplicationsView(PartnerCoordinatorOnly, DetailView):
...
 
+    def dispatch(self, request, *args, **kwargs):
+        partner = self.get_object()
+        auth_method = partner.authorization_method
+        if auth_method == Partner.EMAIL or auth_method == Partner.CODES:
+            return super(SendReadyApplicationsView, self).dispatch(
+                request, *args, **kwargs
+            )
+        else:
+            raise Http404("Application's for this Partner are sent automatically")
```
Samwalton9: Just one small typo - "Application's" should be "Applications".

Samwalton9:
This is looking great! There's just one typo that I've left a comment about, if you can fix that I'll get this merged :)

TWLight/applications/views.py
lalit97: fixed it :)

lalit97:
Thanks, I have fixed the typo :)

Samwalton9:
Great, thanks!

PR #410, opened 2020-02-22
Add 'Back-to-top' button for partners list page

lalit97:
The partners page is quite long now. A 'back to top' button was needed.

Bug: T170113

### DIFF
--- a/TWLight/resources/templates/resources/partner_filter.html
+++ b/TWLight/resources/templates/resources/partner_filter.html
@@ -49,4 +49,11 @@
       </div>
     {% endfor %}
   </div>
+
+  <div>
+    {% comment %}Translators: This is a button to go back on the top of page. {% endcomment %}
+    <a id="back-to-top" href="#" class="btn btn-light btn-sm" title="{% trans 'Back to top' %}">
+      <i class="fa fa-chevron-up"></i>
+    </a>
+  </div>
 {% endblock content %}
--- a/TWLight/static/css/local.css
+++ b/TWLight/static/css/local.css
@@ -5,6 +5,7 @@ BODY
 html {
   position: relative;
   min-height: 100%;
+  scroll-behavior: smooth;
 }
 
 body {
@@ -752,3 +753,24 @@ CLEARING COLUMNS
     clear: left;
   }
 }
+
+/* For 'back to top' button on /partners/ */
+#back-to-top {
+  display: inline-block;
+  border-radius: 4px;
+  position: fixed;
+  bottom: 170px;
+  right: 10px;
+  font-size: 28px;
+  z-index: 1000;
+  background-color: #ff9800;
+  color: #fff;
+  opacity: 1;
+  visibility: visible;
+}
+
+#back-to-top:hover {
+  background-color: #333;
+  color: #fff;
+}
+


Samwalton9:
Could you add a screenshot of what this button looks like, please? :)

lalit97:
back-to-top2

lalit97:
Hello @Samwalton9 I thought you just missed this one, please have a look whenever you are free :)

Samwalton9:
Hey lalit - thanks for the reminder. I'm actually going to wait on reviewing this. We've got a lot of moving pieces being merged in to the tool this month and it might just reduce complications to hold off on merging this PR until then. Sorry about that - we'll get to it in a few weeks :)

lalit97:
Okay, no problem :)

jsnshrmn:
I rebased this so I can check it against our current code base and review.

TWLight/static/css/local.css
```diff
@@ -748,3 +748,28 @@ CLEARING COLUMNS
...
+
+#back-to-top:hover{
+  background-color: #333;
+  color: #fff;
+}
+
+#back-to-top.display{
+  opacity: 1;
+  visibility: visible;
+}
```
jsnshrmn, line 775: files should end with a newline

jsnshrmn:
Also, we've moved to an anonymizing reverse proxy for external assets, so it would be great if you updated the URLs to match what is in #447. Since you had no way of knowing that this change was about to happen, I'd be happy to make that update on your behalf if you don't want to.

TWLight/static/css/local.css
lalit97: added a newline :)

lalit97:
> This occludes the footer navigation in mobile views. That needs to be resolved

Fixed it, now working well in mobile view.

> When I load the page without JavaScript, this feature disappears. Please use JavaScript for progressive enhancement where possible. Since this is just navigation between page anchors, it should work without JavaScript.

removed the javascript part.

lalit97:
> Also, we've moved to an anonymizing reverse proxy for external assets, so it would be great if you updated the URLs to match what is in #447. Since you had no way of knowing that this change was about to happen, I'd be happy to make that update on your behalf if you don't want to.

I have removed the urls, since they were already present in `base.html`. Thank you for such a detailed review. I have fixed all the points you have mentioned :)

TWLight/resources/templates/resources/partner_filter.html
```diff
@@ -49,4 +49,8 @@ <h2>{% trans "Browse Partners" %}</h2>
       </div>
     {% endfor %}
   </div>
+
+  <a id="back-to-top" href="#" class="btn btn-light btn-sm">
```
suecarmol, line 53: Can you add a tooltip that says `Back to top` along with its corresponding translator comment?

TWLight/static/css/local.css
```diff
@@ -752,3 +753,24 @@ CLEARING COLUMNS
     clear: left;
   }
 }
+
+/* For 'back to top' button on /partners/ */
+#back-to-top {
+  display: inline-block;
+  border-radius: 4px;
+  position: fixed;
+  bottom: 210px;
```
suecarmol, line 762: I'm going back and forth on whether this should be lower (around 170px). At 170px, the button doesn't cover the footer in any device I have tested it on. Is there a reason why you chose 210px?

TWLight/resources/templates/resources/partner_filter.html
lalit97: added a tooltip :)

TWLight/static/css/local.css
lalit97: Reduced the value to 170px you are right it is working fine on this value at all devices :)

PR #413, opened 2020-02-25
T234552: Users who are blocked on any project should require manual review for Bundle approval

jsnshrmn:
https://phabricator.wikimedia.org/T234552

Okay, we had to pay a bit of technical debt to add this functionality.

In this PR, I've factored the many implementations of get_global_userinfo out to a single static function and moved it along with all of the recent editor eligibility/validity functions out to a helper module.

We previously were only getting blocked status for the SUL "home" project account rather than global blocks. We're now getting a list of local project accounts with block status for each one as part of global_userinfo. The block check essentially works the same as before, except it's now checking for a block in *any* account attached to the global user.

There is a new Editor.ignore_wp_blocks Boolean property that allows users to be treated as if they had no blocks, even if they do. Basically, this overrides the check for blocks while calculated Editor.is_valid.

On the editor profile, the editor should show as failing the block check, but being valid anyway.

There is currently no UI for setting ignore_wp_blocks, so that will need to be done in admin by a superuser for now.

### DIFF
--- a/TWLight/users/management/commands/user_check.py
+++ b/TWLight/users/management/commands/user_check.py
@@ -1,15 +1,12 @@
-import json
 import logging
-import urllib.request, urllib.error, urllib.parse
 
-from datetime import datetime
-from django.conf import settings
-from django.db import models
 from django.contrib.auth.models import User
-from django.core.management.base import BaseCommand, CommandError
-from ....users.models import Editor, UserProfile
+from django.core.management.base import BaseCommand
+from TWLight.users.helpers.editor_data import editor_global_userinfo
 
 logger = logging.getLogger(__name__)
+
+# TODO: Check if the assertion inside editor_global_userinfo bubbles up to this command.
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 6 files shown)


Samwalton9:
Quick first thought as this relates to the new UI elements - if we're still showing "you're blocked" on the homepage and user profile, we should adjust that if they've been overridden, so we're not still telling them to contact us. I'll open a PR for that after this is merged to staging.

TWLight/users/management/commands/user_check.py
```diff
@@ -1,16 +1,13 @@
...
-from django.db import models
 from django.contrib.auth.models import User
-from django.core.management.base import BaseCommand, CommandError
-from ....users.models import Editor, UserProfile
+from django.core.management.base import BaseCommand
+from TWLight.users.helpers.editor_data import editor_global_userinfo
 
 logger = logging.getLogger(__name__)
 
+# TODO: Check if the assertion inside editor_global_userinfo bubbles up to this command.
```
Samwalton9: Is this still outstanding for this PR, or later?

jsnshrmn: later

PR #428, opened 2020-03-06
support streams to partner consolidation

jsnshrmn:
Adds some signal receivers to handle authorization de-duplication and partner configuration after collection deletion. Needs tests.

### DIFF
--- a/TWLight/applications/__init__.py
+++ b/TWLight/applications/__init__.py
@@ -1,0 +1,1 @@
+default_app_config = "TWLight.applications.app.Config"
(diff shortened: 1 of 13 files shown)


jsnshrmn:
I've added testing for this, but the auth de-duplication assertion is failing despite the fact that the code works as I'd expect in practice. Setting this down for the hope that fresh eyes (either mine another day, or someone else's sooner) will see the issue clearly.

jsnshrmn:
I've spent the day noodling around and haven't gotten to the bottom of the test failures. I could really use a sanity check on this. If you don't see anything jumping out at you in a readthrough, perhaps you could take it for a spin and see if it behaves as desired?

jsnshrmn:
This ended up with a fairly major reworking of our signals, which was on our list of technical debt that needed paying. Refactored signals everywhere except emails app.

PR #429, opened 2020-03-07
Handle modification of application status from INVALID to anything else

lalit97:
If a coordinator tries to modify application status from INVALID to anything else
then return the coordinator back to the app with an error message

Bug: T239503


lalit97:
Video showing an example of changes

T239503

Samwalton9:
This looks perfect! Thanks :)

lalit97:
My pleasure, thanks :)

PR #431, opened 2020-03-13
Add proxy/bundle management command and associated launch email

Samwalton9:
- Added a new management command, `proxy_bundle_launch`.
- Command sends all users a one-off email about the proxy/bundle launch
- Tracks, via a model field, which users we sent the email to, in case we need to restart the command due to an error or somesuch.

We can now add the session-dropping part to the same command (Jason to do)

### DIFF
--- a/TWLight/emails/tasks.py
+++ b/TWLight/emails/tasks.py
@@ -179,12 +179,7 @@ def send_proxy_bundle_launch_notice(sender, **kwargs):
 
     email = ProxyBundleEmail()
 
-    email.send(
-        user_email,
-        {
-            "username": user_wp_username,
-        },
-    )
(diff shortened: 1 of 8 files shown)


Samwalton9:
I'll rebase this to master because it doesn't depend on any proxy/bundle code, then we can get translations.

Samwalton9:
Well that didn't work.

Samwalton9:
That's better. 'Sent' flag still to come.

Samwalton9:
This should be good to go for merge into Master since it doesn't touch any proxy-specific code. Then we can get some translations in in advance.

Samwalton9:
TODO: Do something about `XXX`.

PR #444, opened 2020-04-07
WIP T244896

jsnshrmn:
https://phabricator.wikimedia.org/T244896

Changes partner field on Authorization from foreign key to many-to-many.
The idea here is that each bundle-eligible user would have a single authorization to cover all the resources included in the bundle.

We previously had `unique_together = ("user", "partner", "stream")` for authorizations, but that's not possible with many to many relationships, so we'll need to add back in some constraints at the application layer.

As part of this work, I discovered that we were missing pre_save validation on auths, so I've added that in since we need it.


jsnshrmn:
Status update:

DONE:
- branch currently implements Authorizations with a many to many relationship for Partners.
- I've now resolved all the bugs that were discovered when I turned on validation for authorizations.
- I've added back in some constraints/validation to only allow one non-bundle partner per auth, and to only allow auths to contain partners with the same auth method.

TODO:
- tests
- write out the code to handle the creation and management of bundle auths. signals for updating auths when partners move to bundle etc
- the ezproxy part simply needs to check to see if it's a bundle auth and then return a fixed group code to represent that, rather than looping through each partner.
- data migration to keep from thowing away the current auths

Samwalton9:
This week I'll work to:

- Add tests for the new authorizations helper function
- Verify our authorization tests do all the things we hope they do, especially given the new relations and constraints
- Add the Bundle eligibility authorization creation check to the user login and cron job update code

TWLight/ezproxy/views.py
```diff
@@ -32,6 +32,9 @@ def get(request, url=None, token=None):
         if not username:
             raise SuspiciousOperation("Missing Editor username.")
 
+        if request.user.editor.wp_bundle_eligible:
```
Samwalton9, line 35: Should this be checked with this bool or by looking for the user's Bundle authorization?

jsnshrmn: Yeah, it should check the bundle auth. Those didn't exist when this was written.

Samwalton9:
Above I have:
- Added a new editor function called `update_bundle_authorization` which updates the editors Bundle authorization, creating it if it doesn't exist and expiring it if it previously existed but they are no longer eligible. TODO: Un-expire auths if the user has become eligible again.
- Called this function on user login and in our cron update script
- Added tests for this (TODO: Add a test for the case I missed, identified above)

- Added a pre_save and post_save signal to update bundle auths when partner objects are modified. If a partner is updated from not-bundle to bundle, or not available to available, and vice versa, we now update everyone's auths accordingly.
- Added tests for this.

- Updated the ezproxy/views bundle group code to check the authorization directly rather than using the eligibility data.

TWLight/users/models.py
```diff
@@ -357,9 +367,63 @@ def get_wp_groups_display(self):
...
+        if necessary.
+        The list of partners for the auth will be kept up-to-date
+        elsewhere after initial creation, so no need to worry about
+        updating an existing auth with the latest bundle partner
+        changes here.
+        """
+        # Although we have multiple partners with the BUNDLE authorization
+        # method, we should only ever find one or zero authorizations
+        # for bundle partners.
+        bundle_authorization = Authorization.objects.filter(
```
Samwalton9, line 394: Out of the gate we could grab active authorization(s) so that we do one check for whether it exists or not.

TWLight/users/models.py
```diff
@@ -357,9 +367,63 @@ def get_wp_groups_display(self):
...
+            user=self.user, partners__authorization_method=Partner.BUNDLE
+        ).distinct()  # Required because partners__authorization_method is ManyToMany
```
Samwalton9, line 396: We could reduce db hits by doing something like `distinct().get()` here.

TWLight/users/models.py
```diff
@@ -357,9 +367,63 @@ def get_wp_groups_display(self):
...
+        if bundle_authorization.count() > 1:
+            # This is unexpected and implies something else is wrong.
+            raise MultipleObjectsReturned
+        if bundle_authorization.exists():
```
Samwalton9, line 400: This is unnecessary - count = 1 is the same.

TWLight/users/models.py
```diff
@@ -357,9 +367,63 @@ def get_wp_groups_display(self):
         else:
             return None
 
+    def authorized_for_bundle(self):
```
Samwalton9, line 370: Rename to `wp_bundle_authorized`

TWLight/users/models.py
```diff
@@ -357,9 +367,63 @@ def get_wp_groups_display(self):
...
+        user_bundle_authorizations = Authorization.objects.filter(
```
Samwalton9, line 371: This could just be the filter exist check, it's a boolean already.

TWLight/users/models.py
```diff
@@ -357,9 +367,63 @@ def get_wp_groups_display(self):
...
+            user=self.user, partners__authorization_method=Partner.BUNDLE
+        ).exists()
```
Samwalton9, line 373: This should ensure the auth is valid. Add a test for this too.

Samwalton9:
Final TODO is that I need to reduce some duplicate auth-getting logic.

Samwalton9:
Probably best to look at the overall diff here for today's commits since I did a lot of overwriting of previous changes: https://github.com/WikipediaLibrary/TWLight/compare/aa1d370..e854ffc

jsnshrmn:
Will do. Thanks Sam!

PR #446, opened 2020-04-20
Add a new application field for waitlist status

lalit97:
This is a field that would be True if the application was
filed while the partner was waitlisted, and False otherwise

Bug: T206553


TWLight/resources/tests.py
```diff
@@ -526,6 +526,43 @@ def test_toggle_waitlist_access(self):
...
+
+        # Test partner status is changed to Waitlist
+        self.assertEqual(partner.status, Partner.WAITLIST)
+
+        # Test if waitlist_status is True for all applications
+        # which are Pending or Under Discussion for this partner
+        applications = Application.objects.filter(partner=partner)
+        for app in applications:
+            if app.status in [Application.PENDING, Application.QUESTION]:
+                self.assertEqual(app.waitlist_status, True)
```
Samwalton9, line 564: Instead of filtering the list of all partner applications, then using an if statement to check application status, we could instead filter for those statuses at the outset, like `applications = Application.objects.filter(partner=partner, status__in=[Application.PENDING, Application.QUESTION])`

TWLight/resources/views.py
```diff
@@ -233,6 +233,14 @@ def post(self, request, *args, **kwargs):
             partner.status = Partner.WAITLIST
             # Translators: When an account coordinator changes a partner from being open to applications to having a 'waitlist', they are shown this message.
             msg = _("This partner is now waitlisted")
+
+            # Set waitlist_status to True for all the applications
+            # which are Pending or Under Discussion for this partner
+            applications = Application.objects.filter(partner=partner)
+            for app in applications:
+                if app.status in [Application.PENDING, Application.QUESTION]:
```
Samwalton9, line 241: Likewise, this could be filtered in the same way.

TWLight/resources/tests.py
lalit97: great! this will definitely reduce the time complexity ^_^ 
I have made the changes in the query.

lalit97:
> This looks perfect! I've left two comments, which are just about improving how you filter querysets. Besides that this looks good to me.

Thank you :) I have made changes according to the comments.

> As with the other pull requests, this will need to wait for our big branch merge before we move ahead with approving it.

No problem at all :)

lalit97:
Thank you :)

PR #449, opened 2020-05-11
Staging bundle fixes for my_library

Samwalton9:
We weren't handling the ManyToMany partner relationship on Authorizations in the my_library page. This PR takes a very simple approach to fixing that, both for the Bundle and non-Bundle case.

### DIFF
--- a/TWLight/users/views.py
+++ b/TWLight/users/views.py
@@ -738,7 +738,6 @@ class CollectionUserView(SelfOnly, ListView):
                 if (
                     each_authorization.about_to_expire
                     or not each_authorization.is_valid
-                    or not each_authorization.is_bundle
                 ):
                     latest_app = each_authorization.get_latest_app()
                     if latest_app:
--- a/TWLight/users/views.py
+++ b/TWLight/users/views.py
@@ -757,7 +757,7 @@ class CollectionUserView(SelfOnly, ListView):
                                         Application.QUESTION,
                                         Application.APPROVED,
                                     ),
-                                    partner=each_authorization.partners.first(),
+                                    partner=each_authorization.partners.get(),
                                 ).latest(
                                     "date_created"
                                 )
--- a/TWLight/users/views.py
+++ b/TWLight/users/views.py
@@ -738,6 +738,7 @@ class CollectionUserView(SelfOnly, ListView):
                 if (
                     each_authorization.about_to_expire
                     or not each_authorization.is_valid
+                    or not each_authorization.is_bundle
                 ):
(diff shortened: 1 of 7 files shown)


Samwalton9:
Tested locally with Bundle partners, proxy partners, and non-bundle partners.

We should probably have actual tests for this.

jsnshrmn:
It look like we have a test that should cover this:
https://github.com/WikipediaLibrary/TWLight/blob/staging/TWLight/users/tests.py#L313
but the test is broken. It's looping through an always empty queryset to check the items for problems. Working on a fix

jsnshrmn:
I now have this test at the point where it errors without this PR and fails with it. Will circle back around tomorrow

TWLight/users/views.py
```diff
@@ -756,7 +757,7 @@ def get_context_data(self, **kwargs):
                                         Application.QUESTION,
                                         Application.APPROVED,
                                     ),
-                                    partner=each_authorization.partners,
+                                    partner=each_authorization.partners.first(),
```
jsnshrmn, line 760: seems like this should be `.get()` instead of first so we raise an exception if there is more than one result? Maybe we need to split out some of this to its own method function to allow separate handing for bundle

Samwalton9:
This still isn't pretty, but it works. Because we need information from both the auth and partner objects for each tile and make no judgement on what kind of auth it is until we're in `collection_tile.html` it was hard to disentangle the code in a cleaner way. Fixing the partner/stream issue will go a long way to making this more readable.

I also couldn't find a good way to alphabetise across proxy and bundle tiles because of the way we've set up Bundle authorizations with multiple partners it's hard to pull apart the partners for ordering when you need to retain the overarching authorization information for bundle tiles.

jsnshrmn:
If you want to simplify the template and make sorting easier, you could traverse the partners, streams, and authorizations in the view code and populate a flat list of dicts in the context data, where each dict has the required/merged info for one tile. Does that make sense?

Samwalton9:
I did consider that but I'm not sure it makes the template a huge amount tidier since it's mostly conditionally sorting html/css. Would result in a lot of reorganisation but are there a significant number of places in `collection_tile` that we'd actually cut down on code?

jsnshrmn:
I would expect a significant reduction in those nested if/elif statements would be possible since you could be doing those checks in the view code and assigning some kind of type and status attributes to the dicts. then the template could take a more declarative approach.

jsnshrmn:
we'll punt that cleanup to a separate pr so that bundle ua can continue.

PR #450, opened 2020-05-14
More staging bundle fixes

Samwalton9:
- Users are now redirected to the homepage after agreeing to the Terms of Use
  - Removed a piece of the login code that wasn't being used
  - When a partner is switched to Bundle, we now delete any previously existing authorizations to that partner, since they're no longer valid.
    - Added a test for this, which failed before my signal changes.
  - If we haven't collected Bundle eligibility data for a user yet (as determined by whether we've collected `editor.wp_editcount_prev`), display "Unknown" instead of "No" next to eligibility criteria on application review pages.
  - Instead of looping authorizations and partners in my_library, we now construct a list of dictionaries, one item per tile, bringing some of our logic out into view code.
    - This also fixes the authorization count on the tab when the user has a Bundle authorization.

### DIFF
--- a/TWLight/users/helpers/authorizations.py
+++ b/TWLight/users/helpers/authorizations.py
@@ -48,14 +48,14 @@ def get_valid_partner_authorizations(partner_pk, stream_pk=None):
         return Authorization.objects.none()
 
 
-def sort_authorizations_into_my_library_list(authorizations):
+def sort_authorizations_into_resource_list(authorizations):
     """
     Given a queryset of Authorization objects, return a
     list of dictionaries, sorted alphabetically by partner
     or stream name, with additional data computed for ease
     of display in the my_library template.
     """
-    authorization_list = []
+    resource_list = []
     if authorizations:
         for authorization in authorizations:
             for partner in authorization.partners.all():
@@ -66,8 +66,8 @@ def sort_authorizations_into_my_library_list(authorizations):
                     name_string = stream.name
                 else:
                     name_string = partner.company_name
-                authorization_list.append({"name": name_string})
-                this_item = authorization_list[-1]
+                resource_list.append({"name": name_string})
+                this_item = resource_list[-1]
 
                 this_item["partner"] = partner
                 this_item["authorization"] = authorization
@@ -83,8 +83,8 @@ def sort_authorizations_into_my_library_list(authorizations):
                 this_item['valid_authorization_with_access_url'] = access_url and valid_authorization and authorization.user.userprofile.terms_of_use
 
         # Alphabetise by name
-        authorization_list = sorted(authorization_list, key=lambda i: i["name"])
+        resource_list = sorted(resource_list, key=lambda i: i["name"])
 
-        return authorization_list
+        return resource_list
     else:
         return None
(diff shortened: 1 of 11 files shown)


TWLight/users/helpers/authorizations.py
```diff
@@ -46,3 +46,45 @@ def get_valid_partner_authorizations(partner_pk, stream_pk=None):
...
+
+
+def sort_authorizations_into_my_library_list(authorizations):
+    """
+    Given a queryset of Authorization objects, return a
+    list of dictionaries, sorted alphabetically by partner
+    or stream name, with additional data computed for ease
+    of display in the my_library template.
+    """
+    authorization_list = []
```
jsnshrmn, line 58: this isn't a list of authorizations. I'd name it `my_library_list`

TWLight/users/templates/users/collection_tile.html
```diff
@@ -9,83 +9,67 @@
 {% endcomment %}
 
 <div class="panel panel-default full-width">
-  <div class="panel-body top-border {% if each_authorization.about_to_expire %}expiry-warning{% endif %}">
-    {% if each_authorization.is_bundle %}
+  <div class="panel-body top-border {% if collection.authorization.about_to_expire %}expiry-warning{% endif %}">
```
jsnshrmn, line 12: collection is the user-facing name of streams. Maybe `my_library_resource`? I realize the template is called collection tile, which is also not ideal.

TWLight/users/templates/users/collection_tile.html
```diff
@@ -97,14 +81,14 @@
...
+    {% if not collection.authorization.is_bundle %}
+      {% if collection.authorization.latest_sent_app and not collection.authorization.open_app %}
         {% comment %} Translators: Hovertext for when renewals for a partner is not available and the renew button is disabled. {% endcomment %}
-        <span class="pull-right" {% if not each_partner.renewals_available %}title="{% trans 'Renewals are not available for this partner' %}"{% endif %}>
-          <a href="{% url 'applications:renew' each_authorization.latest_sent_app.pk %}" class="btn btn-default pull-right {% if not each_partner.renewals_available %}disabled{% endif %}" role="button">
-            {% if each_authorization.is_valid %}
+        <span class="pull-right" {% if not collection.partner.renewals_available %}title="{% trans 'Renewals are not available for this partner' %}"{% endif %}>
+          <a href="{% url 'applications:renew' collection.authorization.latest_sent_app.pk %}" class="btn btn-default pull-right {% if not collection.partner.renewals_available %}disabled{% endif %}" role="button">
+            {% if collection.authorization.is_valid %}
```
jsnshrmn, line 91: I feel like there is room to rethink areas like this, but that's very much a code quality backlog kind of thought, not a launch blocker.

TWLight/users/templates/users/my_library.html
```diff
@@ -50,12 +50,10 @@ <h3 class="inline-header">{% trans 'Your Library' %}</h3>
     ">
       <br>
       <div class="row row-flex">
-        {% for each_authorization in proxy_bundle_authorizations %}
-          {% for each_partner in each_authorization.partners.all %}
-            <div class="col-flex col-xs-12 col-sm-4 col-md-3">
-              {% include "users/collection_tile.html" %}
```
jsnshrmn, line 56: the more I look at this, the more I hate `collection_tile`. really should be `resource_tile`. just for code readability since we have collections/streams. again, not really your doing.

TWLight/users/helpers/authorizations.py
jsnshrmn: or `my_resource_list`

PR #451, opened 2020-05-15
Even more Bundle fixes

Samwalton9:
- Coordinators can stop being coordinators in two ways - either they're no longer in the coordinator user group or they delete their data. Both cases were leading to server errors when we tried to do things with the related Partner or Authorization objects.
  - ~~We now no longer require an Authorization's `authorizer` to be a coordinator, because a coordinator might have authorized a user and then stopped being a coordinator.~~
    - Actually, we still do this validation but only when the object is created.
  - When a user deletes their data, if they're a coordinator we need to shift the `authorizer` to TWL Team so that the authorization is still valid (i.e. `is_valid` returns True)
- ~~When an authorization doesn't have a specific stream attached, but the partner has streams, the user should be authorized for all streams and have them show up individually in their library.~~
- Don't send a message after not agreeing to ToU, because we always show a very similar message anyway.
- Reverted my target URL + proxy validation on partner resources. We want to be able to set auth method at the resource level even when target URLs are only configured on streams.
- I had excluded Waitlist status from our resource signal considerations when updating Bundle authorizations. That didn't make much sense because we needed to update auths if, for example, switching a previously waitlisted partner to available + bundle in one go. We now basically consider waitlist to be synonymous with not_available for those purposes.

### DIFF
--- a/TWLight/users/helpers/authorizations.py
+++ b/TWLight/users/helpers/authorizations.py
@@ -53,7 +53,7 @@ def create_resource_dict(name, authorization, partner, stream):
         "name": name,
         "partner": partner,
         "authorization": authorization,
-        "stream": stream
+        "stream": stream,
     }
 
     if stream:
@@ -64,13 +64,12 @@ def create_resource_dict(name, authorization, partner, stream):
 
     valid_authorization = authorization.is_valid
     resource_item["valid_proxy_authorization"] = (
-            partner.authorization_method == partner.PROXY
-            and valid_authorization
+        partner.authorization_method == partner.PROXY and valid_authorization
     )
     resource_item["valid_authorization_with_access_url"] = (
-            access_url
-            and valid_authorization
-            and authorization.user.userprofile.terms_of_use
+        access_url
+        and valid_authorization
+        and authorization.user.userprofile.terms_of_use
     )
 
     return resource_item
@@ -88,9 +87,7 @@ def sort_authorizations_into_resource_list(authorizations):
         for authorization in authorizations:
             for partner in authorization.partners.all():
                 stream = authorization.stream
-                partner_streams = Stream.objects.filter(
-                    partner=partner
-                )
+                partner_streams = Stream.objects.filter(partner=partner)
                 if partner_streams and not stream:
                     # If this authorization wasn't to a specific stream, but the
                     # partner has streams, we assume that means all streams
@@ -100,10 +97,8 @@ def sort_authorizations_into_resource_list(authorizations):
                     for stream in partner_streams:
                         resource_list.append(
                             create_resource_dict(
-                                stream.name,
-                                authorization,
-                                partner,
-                                stream)
+                                stream.name, authorization, partner, stream
+                            )
                         )
                 else:
                     # Name this item according to whether this authorization is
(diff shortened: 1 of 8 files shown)


Samwalton9:
> Create and update can have different validation requirements

This makes sense, hadn't considered that. I assume to do this we'd need to move what was in `limit_choices_to` to a validation function that's only called on object creation?

> This really breaks the auth - stream model

What in particular breaks it?

TWLight/users/helpers/authorizations.py
```diff
@@ -59,35 +86,33 @@ def sort_authorizations_into_resource_list(authorizations):
     if authorizations:
         for authorization in authorizations:
             for partner in authorization.partners.all():
-                # Name this item according to whether this authorization is
-                # to a stream
                 stream = authorization.stream
-                if stream:
-                    name_string = stream.name
+                partner_streams = Stream.objects.filter(partner=partner)
+                if partner_streams and not stream:
```
jsnshrmn, line 91: If a user is authorized to access a stream, they should have an authorization with that stream in it.

Samwalton9: Oh, yep. Fixing this the 'proper' way would mean going through the whole `partner`->`partners` thing with streams too, right?

The alternative being we always split partners up into their respective streams if they're switching to Bundle?

Samwalton9: I think the latter would only need to be done for Oxford University Press - it's the only partner with streams that's moving to Bundle.

jsnshrmn: In that case, the easiest thing to do would be to just make a new OUP bundle partner, and mark the current one as unavailable.

Samwalton9: Oh that's a good point, since we don't care about existing apps/auths. I think it's something like 10 collections, which means nearly half of the Library Bundle will be individual OUP-published sources, but this may be the best option.

jsnshrmn: FYI, our current OUP ezproxy config:
https://github.com/WikipediaLibrary/twlight_ezproxy/blob/19dd11e4f6947abdd90699ffda536117a83a2aac/ezproxy.cfg#L257-L272

jsnshrmn: If we wanted to split it out, each one of those oxford entries could be a partner.

jsnshrmn: Or they could all be a single partner

Samwalton9: Will clarify the full list with Nikki,

Samwalton9: Nikki's going to move forward with creating new partners for OUP. I'll revert this change and we'll move ahead with not considering Bundle streams at all.

Samwalton9: Oh and she's going to clarify the list of collections so that we can update the proxy config.

Samwalton9:
Ok, I reverted the Bundle/stream change per our conversation about Oxford University Press streams being shifted to individual collections so we don't need to worry about this anymore.

Second, I added validation for the Authorization model but as part of the `clean()` functionality instead of being directly on the model. This allows us to verify the coordinator's user groups on object creation and ignore them on update. Does that seem a sensible improvement?

TWLight/users/helpers/authorizations.py
```diff
@@ -2,7 +2,7 @@
 
 from django.db.models import Q
 
-from TWLight.resources.models import Partner
+from TWLight.resources.models import Partner, Stream
```
jsnshrmn, line 5: unused import
```suggestion
from TWLight.resources.models import Partner
```

PR #468, opened 2020-06-23
Admin interface validation for target URL uniqueness

uyscuti-wiki:
[T247015](https://phabricator.wikimedia.org/T247015)


suecarmol:
💭 This looks good, but I wonder if we need to add integration tests since this helper function is called in the `clean` function. Maybe we should check if the form or API that calls this model actually shows the `ValidationErrors` that are thrown by this function.

PR #472, opened 2020-07-01
T255167: Additional CI check for Python translation style requirements

jsnshrmn:

## Description

- Added perl package to Dockerfile
- Created perl script for checking individual files for issues (python checks complete, html jinja template checks incomplete)
- Call perl script from existing test script (only checks `.py` files, not `.html` files)

## Rationale

From the ticket:
>Django requires particular formatting when it comes to translatable strings, particularly:
>
>- Translator comments must be immediately above the corresponding string
>- Translated strings can't be preceded by a newline, eg
>
>Our code linter (black) tends to cause breakages here, but contributors may also introduce these errors directly.
>
>We can add an additional script to our CI checks, after black has run, to ensure that both conditions are met.

## Phabricator Ticket

https://phabricator.wikimedia.org/T255167

## How Has This Been Tested?

- Added new top level tests directory
- Added shunit2 for shell unit tests
- Created test data for the perl script (python)
- Created unit tests for the perl script (python)

## Types of changes


jsnshrmn:
I'm very open to feedback on the layout / design / implementation of the shell tests. Very much something I threw together in the service of making sure this new script works.

jsnshrmn:
I chose shunit2 for tests here as it was an easy way to get started since installation meant dropping a single shell file into source. Most of the effort has gone into developing the test data and writing my regular expressions appropriately.
I'd be happy to switch over to something like [shellspec](https://github.com/shellspec/shellspec) that supports coverage checks along with a bunch of other features if we like that better.

bin/virtualenv_test.sh
```diff
@@ -15,8 +15,19 @@ set -euo pipefail
         echo "black --target-version py37 --check TWLight"
         if black --target-version py37 --check TWLight
         then
+            echo "${TWLIGHT_HOME}/tests/shunit/twlight_i18n_lint_test.sh"
+            ${TWLIGHT_HOME}/tests/shunit/twlight_i18n_lint_test.sh
+
+            # https://phabricator.wikimedia.org/T255167
```
Samwalton9, line 21: We might want to link to https://github.com/WikipediaLibrary/TWLight/wiki/Translation instead of this Phab task?

tests/shunit/data/bad_i18n_comment_1.py
```diff
@@ -0,0 +1,3 @@
+
```
suecarmol, line 1: Although the title of the file explains that this is a bad comment, I think it a comment explaining why this example is wrong would be a good idea. This comment should be added to all the `tests/shunit/data` files.

tests/shunit/twlight_i18n_lint_test.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+
+shopt -s expand_aliases
+
+testBadPyNewlines() {
+    start=1
+    end=6
```
suecarmol, line 7: Since the end variable is fixed, I think we should add a comment to mention that if you add a new test case, you need to change this variable so that it is run in the script.

tests/shunit/twlight_i18n_lint_test.sh
```diff
@@ -0,0 +1,35 @@
...
+    for ((i=$start; i<=$end; i++))
+    do
+        file=${TWLIGHT_HOME}/tests/shunit/data/bad_i18n_newline_$i.py
+        assertFalse "${file} should cause an error." "perl ${TWLIGHT_HOME}/bin/twlight_i18n_lint.pl ${file}"
+    done
+}
+
+testGoodPyComments() {
+    start=1
+    end=2
```
suecarmol, line 17: Same here

tests/shunit/twlight_i18n_lint_test.sh
```diff
@@ -0,0 +1,35 @@
...
+    for ((i=$start; i<=$end; i++))
+    do
+        file=${TWLIGHT_HOME}/tests/shunit/data/good_i18n_comment_$i.py
+        assertTrue "${file} should not cause an error." "perl ${TWLIGHT_HOME}/bin/twlight_i18n_lint.pl ${file}" ||:
+    done
+}
+
+testBadPyComments() {
+    start=1
+    end=2
```
suecarmol, line 27: Same

bin/virtualenv_test.sh
jsnshrmn: agreed

tests/shunit/data/bad_i18n_comment_1.py
jsnshrmn: done

tests/shunit/twlight_i18n_lint_test.sh
jsnshrmn: I rewrote this to use a pattern instead of a fixed length loop.

tests/shunit/twlight_i18n_lint_test.sh
jsnshrmn: I rewrote this to use a pattern instead of a fixed length loop

tests/shunit/twlight_i18n_lint_test.sh
jsnshrmn: I rewrote this to use a pattern instead of a fixed length loop

jsnshrmn:
on the 'good' test data: the existing files actually have good comments and newlines, otherwise they wouldn't pass as expected. There are no 'good' multiline string examples.
I've renamed the 'good' test data files to be more logical.

PR #476, opened 2020-07-09
Lutrome - T242196 - Reword 'I accept' button to 'Submit' on Terms of Use page

Lutrome:
[//]: # (Thank you for uploading a PR to the Wikipedia Library!)

## Description
    - Change "I Accept" on TWLight\users\forms.py#L185-186 to "Submit".
    - Change "I Accept" on TWLight\users\templates\users\forms.html#L300 to "Submit".
    - Change "I Accept" on all TWLight\locale\*\django.po files.

## Rationale
[//]: The fact that there is both a checkbox and "I Accept" button on the Terms of Use page is confusing to users, so it makes more sense to change the "I accept" text in the ToU and in the button to "Submit", making it more clear to the user what it does, and the users will understand that the checkbox has to be checked first before clicking the button.
I also changed "I Accept" to "Submit" across the translator comments in all locale files so the translators can begin working on the change as well.

## Phabricator Ticket
[//]: # https://phabricator.wikimedia.org/T242196

## How Has This Been Tested?
[//]: No tests have been added or modified.
[//]: I tested this change by changing the text "I Accept" to "Submit" in the terms.html and forms.py file, then going through the account creation process to confirm the button works as intended.

## Types of changes
What types of changes does your code introduce? Add an `x` in all the boxes that apply:
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

PR #478, opened 2020-07-09
Re-word the button to 'Submit' 

saloniig:
## Why these changes
It's confusing to users that the form on the Terms of Use page has both a check box and an "I Accept" button. Users expect that the button is sufficient to accept, without the check box. Additionally, users can un-accept the terms, to do which they need to un-check the box and then click... "I Accept".
I re-word the button to 'Submit' so that it's more obvious the check box should be selected first.

## Phabricator Ticket
[Phabricator Ticket](https://phabricator.wikimedia.org/T242196)
 
## Types of changes


suecarmol:
Thank you for contributing to the Wikipedia Library, @saloniig! This PR looks good, but we have merged #476 that made the same changes recently. Therefore, I will go ahead and close the PR.

PR #492, opened 2020-07-29
Add staff to be eligible for library bundle

suecarmol:

## Description
Added the ability for all Wikipedia Library staff to access the Library Bundle from the management command `user_update_eligibility`.

## Rationale
Since Wikipedia Library staff need to test the application's functionality, it is crucial that they have access to the library bundle so they can test it.

## Phabricator Ticket
[T252429](https://phabricator.wikimedia.org/T252429)

## How Has This Been Tested?
Automatic tests have been added.

N/A

## Types of changes

### DIFF
--- a/TWLight/users/helpers/editor_data.py
+++ b/TWLight/users/helpers/editor_data.py
@@ -184,8 +184,11 @@ def editor_recent_edits(
     )
 
 
-def editor_bundle_eligible(wp_valid, wp_enough_recent_edits):
-    if wp_valid and wp_enough_recent_edits:
+def editor_bundle_eligible(editor):
+    enough_edits_and_valid = editor.wp_valid and editor.wp_enough_recent_edits
+    # Staff and superusers should be eligible for bundles for testing purposes
+    user_staff_or_superuser = editor.user.is_staff or editor.user.is_superuser
+    if enough_edits_and_valid or user_staff_or_superuser:
         return True
     else:
         return False
--- a/TWLight/users/management/commands/user_update_eligibility.py
+++ b/TWLight/users/management/commands/user_update_eligibility.py
@@ -40,6 +40,7 @@ class Command(BaseCommand):
         if options["datetime"]:
             wp_editcount_updated = datetime.fromisoformat(options["datetime"])
 
+        # Getting all editors that are currently eligible or are staff or are superusers
         editors = Editor.objects.filter(wp_bundle_eligible=True)
         for editor in editors:
             if options["global_userinfo"]:
@@ -49,7 +50,12 @@ class Command(BaseCommand):
                     editor.wp_username, editor.wp_sub, True
                 )
             if global_userinfo:
-                editor.wp_editcount_prev_updated, editor.wp_editcount_prev, editor.wp_editcount_recent, editor.wp_enough_recent_edits = editor_recent_edits(
+                (
+                    editor.wp_editcount_prev_updated,
+                    editor.wp_editcount_prev,
+                    editor.wp_editcount_recent,
+                    editor.wp_enough_recent_edits,
+                ) = editor_recent_edits(
                     global_userinfo["editcount"],
                     editor.wp_editcount_updated,
                     editor.wp_editcount_prev_updated,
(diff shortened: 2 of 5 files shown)


jsnshrmn:
I was imagining this as an override that staffers could toggle on/off to temporarily force bundle eligibility. This approach works too, and is a good way to start simple. Since we're adding all staff here, we should also add superusers, since it's been our practice that superusers get everything that's authorized in a role-based way.

suecarmol:
> I was imagining this as an override that staffers could toggle on/off to temporarily force bundle eligibility. This approach works too, and is a good way to start simple. Since we're adding all staff here, we should also add superusers, since it's been our practice that superusers get everything that's authorized in a role-based way.

Thanks for the input! I have added superusers to the bundle authorization as well.

suecarmol:
> Looking through this more closely:
> it seems like staff / superuser check should be made in `TWLight.users.helpers.editor_data`.
> The current code should work fine for the cron task, but we also check eligibility at login. I would expect the check at login time to potentially overwrite the value from the cron run.

Before I start working on that, I think it would be best to merge #494 so I can make those changes based on that refactoring.

suecarmol:
After much thinking instead of adding the condition on `TWLight.users.helpers.editor_data`, I decided to add it to `TWLight.users.models.Editor.update_from_wikipedia` and keep it in the `update_user_eligibility` command. Let me know what you think!

jsnshrmn:
I think this is over-complicating things. We've decided that staff and superusers are eligible for the bundle, and we have a helper function whose sole purpose is determining bundle eligibility. Let's put the logic there. The end state (for the session and what's stored in the db) will be the same, and the code will be more maintainable.

jsnshrmn:
Since this will require changing the signature of `editor_bundle_eligible` to include  a few user (not editor) fields, you could feed it a whole user object instead of the individual fields.

TWLight/users/tests.py
```diff
@@ -1010,28 +1010,12 @@ def test_is_user_valid(self):
         )
         self.assertTrue(valid)
 
-    def test_editor_eligibility_functions(self):
```
suecarmol, line 1013: @jsnshrmn I removed this test function because I think we already test `editor_bundle_eligible` throughout different tests, including the one that I added in this PR. Let me know if you think I should refactor these tests instead of deleting them

suecarmol:
> I think this is over-complicating things. We've decided that staff and superusers are eligible for the bundle, and we have a helper function whose sole purpose is determining bundle eligibility. Let's put the logic there. The end state (for the session and what's stored in the db) will be the same, and the code will be more maintainable.

You are right, I did overthink this solution. I have changed it. Let me know what you think!

TWLight/users/management/commands/user_update_eligibility.py
```diff
@@ -40,7 +41,12 @@ def handle(self, *args, **options):
         if options["datetime"]:
             wp_editcount_updated = datetime.fromisoformat(options["datetime"])
 
-        editors = Editor.objects.filter(wp_bundle_eligible=True)
+        # Getting all editors that are currently eligible or are staff or are superusers
```
jsnshrmn, line 44: I would revert this change and use a simple migration to set all staff + superusers to eligible.

TWLight/users/helpers/editor_data.py
```diff
@@ -184,8 +184,11 @@ def editor_recent_edits(
     )
 
 
-def editor_bundle_eligible(wp_valid, wp_enough_recent_edits):
-    if wp_valid and wp_enough_recent_edits:
+def editor_bundle_eligible(editor):
```
jsnshrmn, line 187: This is a much better signature than what I did. Thanks!

TWLight/users/tests.py
jsnshrmn: Yep, we have some redundancy there. The deletions are a-okay. We can just leave it at that since we have a big test redesign in our future.

TWLight/users/management/commands/user_update_eligibility.py
suecarmol: Got it! I am removing some test cases I added, since this will break them (but I don't think we'll have an actual problem in production because of the migration I created.

PR #494, opened 2020-07-30
 Refactor editor_global_userinfo

suecarmol:

## Description
Refactor editor_global_userinfo so that it also queries by wp_sub when the query by username does not return a user's information.

Add the username to the message when a user is not found.

## Rationale
If a Wikipedia user changes their username, whenever the Wikipedia Library queries for that user with their old username, the user is never found. By adding a query of the user by their id, we can guarantee a user is returned. The only way a user is not returned is if that account has been deleted. 

## Phabricator Ticket
[T258613](https://phabricator.wikimedia.org/T258613)

## How Has This Been Tested?
Since this function is run whenever a user logs in, I tried logging in and out with different accounts. There are also tests that cover this function. 

N/A

## Types of changes

### DIFF
--- a/TWLight/users/helpers/editor_data.py
+++ b/TWLight/users/helpers/editor_data.py
@@ -58,7 +58,7 @@ def editor_global_userinfo(
             assert wp_sub == global_userinfo["id"]
     except (KeyError, AssertionError):
         global_userinfo = None
-        logger.exception("Could not fetch global_userinfo for User.")
+        logger.exception(f"Could not fetch global_userinfo for User {guiuser}")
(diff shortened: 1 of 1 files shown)


suecarmol:
> This looks pretty good to me, although I do have a (non-blocking) style nitpick. You're using f-strings here, which is of course fine, but were using `.format()` everywhere else in the app. I'd prefer we stuck with the current style unless we decide to refactor to use f-strings everywhere.

Done! I am more partial to f-strings, but since the codebase is using `.format()`, I'll use that from now on

PR #501, opened 2020-08-07
Delete bundle authorizations when users delete their data

suecarmol:

## Description
Bundle authorizations were not being deleted when a user deleted their data. This caused authorizations to be without an attached user.

## Rationale
A bundle authorization without an attached user is of no use in the database, so it would be best if we deleted it.

## Phabricator Ticket
[T258380](https://phabricator.wikimedia.org/T258380)

## How Has This Been Tested?
Added a unit test

## Types of changes

### DIFF
--- a/TWLight/users/views.py
+++ b/TWLight/users/views.py
@@ -562,6 +562,11 @@ class DeleteDataView(SelfOnly, DeleteView):
             user_authorization.date_expires = date.today() - timedelta(days=1)
             user_authorization.save()
 
+        # Delete any bundle authorizations.
+        bundle_auths = user.editor.get_bundle_authorization
+        if bundle_auths:
+            bundle_auths.delete()
+
         # Did the user authorize any authorizations?
         # If so, we need to retain their validity by shifting
         # the authorizer to TWL Team
--- a/TWLight/users/migrations/0062_delete_hanging_userless_bundle_auths.py
+++ b/TWLight/users/migrations/0062_delete_hanging_userless_bundle_auths.py
@@ -1,0 +1,16 @@
+from django.db import migrations
+
+
+def delete_hanging_bundle_auths(apps, schema_editor):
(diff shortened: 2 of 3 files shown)


TWLight/users/views.py
```diff
@@ -562,6 +562,12 @@ def delete(self, request, *args, **kwargs):
             user_authorization.date_expires = date.today() - timedelta(days=1)
             user_authorization.save()
 
+        # Delete bundle authorizations, if user is eligible for bundle
+        if user.editor.wp_bundle_authorized:
```
Samwalton9, line 566: As I understand it, this is a bit of a duplicate check - `wp_bundle_authorized` returns True if a bundle authorization exists (also using `get_bundle_authorization`), and then checks that the authorization `is_valid`. But in this context I don't think we actually care if the auth is valid or not (though it always should be), we'd want to delete it either way. So I think we can just jump straight to seeing if we get an object from `get_bundle_authorization` or not.

TWLight/users/views.py
```diff
@@ -562,6 +562,12 @@ def delete(self, request, *args, **kwargs):
...
+            bundle_auths = user.editor.get_bundle_authorization
+            if bundle_auths is not None:
+                bundle_auths.delete()
+
```
jsnshrmn, line 570: Agree with @Samwalton9 , something like this should do just fine.
```suggestion
        # Delete any bundle authorizations.
        bundle_auths = user.editor.get_bundle_authorization
            if bundle_auths:
                bundle_auths.delete()

```

TWLight/users/views.py
suecarmol: I've changed the code to exclude that condition!

TWLight/users/views.py
suecarmol: Done!

PR #513, opened 2020-08-19
Validation

saloniig:

## Description
Add a new if statement to model's clean function which, if the object has the CODES or LINK access methods, verifies that it also has user_instructions data. If it doesn't, a ValidationError should be raised stating "Partners with automatically sent messages require user instructions to be entered".
## Rationale
In the admin interface, users shouldn't be allowed to save a partner with the Codes or Link access methods which has no  user_instructions  data.

## Phabricator Ticket
[Phabricator Ticket](https://phabricator.wikimedia.org/T259867)

## How Has This Been Tested?

No tests have been added or modified.
The changes can be tested manually. We can go to admin panel and can add the partner and see the changes. When we try to save a partner without adding user instructions,validation error message is received and it fails as expected. When we try to save partner with adding user instruction it works successfully.

[Screenshot](https://gist.github.com/saloniig/3c1e3f7dc91139293e0b5eff6730d3db)

## Types of changes


jsnshrmn (changes requested):
Hi there! Thanks for your contribution, and thanks for waiting patiently for a review. The code change looks good, but there should probably be a test that verifies that this works as expected. Basically, you should have a test that tries to save a partner that fails this check and assert that it fails as expected, should should also try to save that partner after adding user instructions and verify that it works.

 You can see how we tested for another validation here:
https://github.com/WikipediaLibrary/TWLight/blob/31a1137f4c107661f5d03e9cf3a075b3874c11ea/TWLight/resources/tests.py#L427

Let me know if you need help devising a test for this.

saloniig:
@jsnshrmn Testing has been done and it is now working as expected. I have attached screenshots also. In the if condition braces has been added as it was not picking right conditions.

jsnshrmn (changes requested):
Hi there,

This really needs a unit test in `TWLight/resources/tests.py` so that we can automatically test the code as we make other changes to the project. Let me know if you need help with that.

saloniig:
Hi there,
@jsnshrmn I have wriiten the unit test but I am not able to undestand how to use it. I have tested it locally by creating object and it is working fine. Please guide me.

jsnshrmn:
When we talk about writing unit tests, we mean writing the code for automated unit tests, since we have an automated unit test framework. `test_helper_function_for_target_url_uniqueness` I linked to earlier is a method of the `PartnerModelTests` class. I In short, you can add another method to the end of that class to test this code. `test_helper_function_for_target_url_uniqueness`  does most of the steps you'll need for your test, which is to say that it creates new partners with problems and then asserts that they fail validation. Let me know if you need more guidance.

jsnshrmn (changes requested):
Tests work with assertions. So, instead of printing errors, assert the state of the partner, the content of the error message, or the truthiness/falseness of something. Check the example code again to see how we assert good and erroneous partner states and try to make your tests more like that.

TWLight/resources/tests.py
```diff
@@ -474,6 +474,33 @@ def test_helper_function_for_target_url_uniqueness(self):
...
+        partner2.requested_access_duration = (
+            True  # We don't want the ValidationError from requested_access_duration
+        )
+
+        error_msg = "Partners with automatically sent messages require user instructions to be entered"
+
+        try:
+            partner1.clean()
+        except ValidationError as e:
+            self.assertEqual([error_msg], e.messages)
```
jsnshrmn, line 496: Since you have a user instruction defined, there shouldn't be an exception to catch here, right?
Why not run partner1.clean() and assert that it returns the user instruction as expected?

jsnshrmn (changes requested):
Thanks for your work! This is looking better. I found one issue in your test, but this is looking good otherwise!

TWLight/resources/tests.py
```diff
@@ -474,6 +474,30 @@ def test_helper_function_for_target_url_uniqueness(self):
...
+            True  # We don't want the ValidationError from requested_access_duration
+        )
+        partner2.user_instructions = ""
+        partner2.requested_access_duration = (
+            True  # We don't want the ValidationError from requested_access_duration
+        )
+
+        error_msg = "Partners with automatically sent messages require user instructions to be entered"
+
+        self.assertEqual(partner1.clean(), None)
```
jsnshrmn, line 493:
```suggestion

        # This partner should validate without errors.
        self.assertEqual(partner1.clean(), None)
```

TWLight/resources/tests.py
```diff
@@ -474,6 +474,30 @@ def test_helper_function_for_target_url_uniqueness(self):
...
+
+        self.assertRaises(ValidationError, partner2.clean)
+        try:
+            partner2.clean()
+        except ValidationError as e:
+            self.assertEqual([error_msg], e.messages)
+
```
jsnshrmn, line 500:
```suggestion

        # This partner should fail validation.
        self.assertRaises(ValidationError, partner2.clean)
        try:
            partner2.clean()
        except ValidationError as e:
            # We're making sure that it fails because of a lack of user instructions, specifically.
            self.assertEqual([error_msg], e.messages)

```

jsnshrmn (changes requested):
The code and the test look great. I'd only ask that you add some comments. I've added suggestions in those spots, so you can just commit the suggestions to make this ready to merge.

jsnshrmn (approved):
Looks good to me. Thanks for your contribution!

PR #529, opened 2020-09-18
Show recently filed applications and their date

lalit97:
## Description
Show all applications filed in last 3 months and their submission date, on application evaluation page.

## Rationale
Before this we used to show 5 recent applications on evaluation page. But we need to display more applications on that page so coordinators can decide if a user is applying indiscriminately.

## Phabricator Ticket
[T262904](https://phabricator.wikimedia.org/T262904)

## How Has This Been Tested?
I have added 2 new tests for this behaviour. Also removed a test which was previously used to test rendering of 5 recent applications only.
This change can be tested manually by going on application evaluation page and checking recent applications section.

## Screenshots of your changes:
Screen Shot 2020-09-18 at 6 31 53 PM

## Types of changes

### DIFF
--- a/TWLight/applications/templates/applications/application_evaluation.html
+++ b/TWLight/applications/templates/applications/application_evaluation.html
@@ -218,25 +218,46 @@
     {% endif %}    
   </div>
 
-  {% comment %}Translators: This is the title of the section of an application page which contains the Recent Applications made by the applicant{% endcomment %}
-  <h2>{% trans 'Recent Applications' %}</h2>
+  {% comment %}Translators: This is the title of the section of an application page which contains the Recent Applications made by the applicant in the last 90 days.{% endcomment %}
+  <h2>{% trans 'Recent Applications (last 90 days)' %}</h2>
   
-  <div class="row">
-    {% for app in recent_apps %}
-      <div class="col-xs-12 col-sm-4">
-        {% if app.partner.coordinator == request.user or app.partner.coordinator == app.editor.user %}
-          <a href="{{ app.get_absolute_url }}">
-            <strong>{{ app.partner }}</strong>
-          </a>
-        {% else %}
-          <strong>{{ app.partner }}</strong>
-        {% endif %}
-      </div>
-      <div class="col-xs-12 col-sm-8">
-        <span class="label label{{ app.get_bootstrap_class }} label-lg">{{ app.get_status_display }}</span>
-      </div>
-    {% endfor %}
-  </div>
+  {% if recent_apps|length > 0 %}
+    <div class="table-responsive">
+      <table class="table table-hover">
+        <thead>
+          <tr>
+            <th scope="col">Partner</th>
+            <th scope="col">Status</th>
+            <th scope="col">Date Created</th>
+          </tr>
+        </thead>
+        <tbody>
+          {% for app in recent_apps %}
+          <tr>
+            <td>
+              {% if app.partner.coordinator == request.user or app.partner.coordinator == app.editor.user %}
+              <a href="{{ app.get_absolute_url }}">
+                <strong>{{ app.partner }}</strong>
+              </a>
+              {% else %}
+                <strong>{{ app.partner }}</strong>
+              {% endif %}
+            </td>
+            <td>
+              <span class="label label{{ app.get_bootstrap_class }} label-lg">{{ app.get_status_display }}</span>
+            </td>
+            <td>
+              <span>{{ app.date_created }}</span>
+            </td>
+          </tr>
+          {% endfor %}
+        </tbody>
+      </table>
+    </div>
+  {% else %}
+    {% comment %}Translators: This text is shown to Coordinators when applicant has not made any application in the last 90 days. {% endcomment %}
+    <i>{% trans "No applications were made in last 90 days by the applicant" %}</i>
+{% endif %}
 
   {% comment %}Translators: This is the title of the section of an application page which contains the comment thread between the user and account coordinator.{% endcomment %}
   <h2>{% trans 'Discussion' %}</h2>
--- a/TWLight/applications/views.py
+++ b/TWLight/applications/views.py
@@ -9,6 +9,7 @@ import urllib.error
 import urllib.parse
 import urllib.request
 from urllib.parse import urlparse
+from datetime import datetime, timedelta
 
 import bleach
 from crispy_forms.helper import FormHelper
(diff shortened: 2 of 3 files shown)


lalit97:
I have broke down the task of showing recent applications in following 2 ways 

(1) If there are more than 5 applications present in the period of last 3 months than I thought it will be good if we display all of them instead of just limiting to 5 applications. Because that will give coordinators a more clear picture of if an editor has applied indiscriminately.

(2) If there are less than 5 applications present in the period of last 3 months than I thought it will be good if we add some more applications from past i.e. applications which are older than 3 months time period. So that we can show at least 5 recent applications. 
But if there were less than 5 total applications made by an editor than we have to show less than 5 recent applications.

Also the diff of `applications/tests.py` is not showing properly I have tried to make separate commits and to make the diff look better by deleting the previous test and then adding the 2 new tests. But they are mixing somehow.

Samwalton9:
> If there are more than 5 applications present in the period of last 3 months than I thought it will be good if we display all of them instead of just limiting to 5 applications. Because that will give coordinators a more clear picture of if an editor has applied indiscriminately.

I think this makes sense to me, good suggestion.

> If there are less than 5 applications present in the period of last 3 months than I thought it will be good if we add some more applications from past i.e. applications which are older than 3 months time period. So that we can show at least 5 recent applications.

I think this, however, over-complicates things. We should show coordinators as little information as necessary to make a decision - seeing applications made more than 3 months ago doesn't, in my opinion, help them make a decision. So I think it should only show last-3-months applications in this case.

> Also the diff of applications/tests.py is not showing properly I have tried to make separate commits and to make the diff look better by deleting the previous test and then adding the 2 new tests. But they are mixing somehow.

One of the engineers can confirm, but I think this is just a quirk of how Github is calculating the diff, I wouldn't worry :)

lalit97:
> I think this makes sense to me, good suggestion.

thanks :)

> I think this, however, over-complicates things. We should show coordinators as little information as necessary to make a decision - seeing applications made more than 3 months ago doesn't, in my opinion, help them make a decision. So I think it should only show last-3-months applications in this case.

Okay got your point will make changes according to this. Also I have a design query if there are 0 applications found in last-3-months then should we show an empty table in Recent Applications sections or should we do not show the whole Recent Applications section(i.e. not even the heading) or maybe we can show a message below the heading that no applications are made recently. Which one should I pick for this case?

> One of the engineers can confirm, but I think this is just a quirk of how Github is calculating the diff, I wouldn't worry :)

okay if you or any other member does not face any difficulty while reviewing the changes then it is okay, otherwise I will try to come up with some solution if there is any :)

Samwalton9:
> Also I have a design query if for an editor there are 0 applications found in last-3-months then should we show an empty table in Recent Applications sections or should we do not show the whole Recent Applications section(i.e. not even the heading) or maybe we can show a message below the heading that no applications are made recently. Which one should I pick for this case?

In this case I think it would be best to show no table or information.

TWLight/applications/templates/applications/application_evaluation.html
```diff
@@ -221,21 +221,41 @@ <h2>{% trans 'Application' %}</h2>
...
-          <strong>{{ app.partner }}</strong>
-        {% endif %}
-      </div>
-      <div class="col-xs-12 col-sm-8">
-        <span class="label label{{ app.get_bootstrap_class }} label-lg">{{ app.get_status_display }}</span>
-      </div>
-    {% endfor %}
+  <div class="table-responsive">
+    <table class="table table-hover table-bordered">
+      <thead style="background-color: #e3f2fd;">
```
suecarmol, line 226: I think no `background-color` on this header would be better.

TWLight/applications/templates/applications/application_evaluation.html
```diff
@@ -221,21 +221,41 @@ <h2>{% trans 'Application' %}</h2>
...
+        <tr>
+          <th scope="col">SN</th>
```
suecarmol, line 228: I would also remove this column. I don't think it really adds any relevant information.

TWLight/applications/templates/applications/application_evaluation.html
```diff
@@ -221,21 +221,41 @@ <h2>{% trans 'Application' %}</h2>
   {% comment %}Translators: This is the title of the section of an application page which contains the Recent Applications made by the applicant{% endcomment %}
   <h2>{% trans 'Recent Applications' %}</h2>
```
suecarmol, line 222: Should we change this to `Recent Applications in the Last 90 Days`, or should we mention that the applications listed here were made in the last 90 days? @lalit97 @Samwalton9 @jsnshrmn your thoughts about this would be appreciated :)

TWLight/applications/templates/applications/application_evaluation.html
```diff
@@ -221,21 +221,41 @@ <h2>{% trans 'Application' %}</h2>
...
-        {% else %}
-          <strong>{{ app.partner }}</strong>
-        {% endif %}
-      </div>
-      <div class="col-xs-12 col-sm-8">
-        <span class="label label{{ app.get_bootstrap_class }} label-lg">{{ app.get_status_display }}</span>
-      </div>
-    {% endfor %}
+  <div class="table-responsive">
+    <table class="table table-hover table-bordered">
```
suecarmol, line 225: I feel that a table without borders looks better. Could you please remove the `table-bordered` class?

TWLight/applications/views.py
```diff
@@ -960,13 +961,35 @@ def get_context_data(self, **kwargs):
...
+        latest_recent_apps = (
+            Application.objects.filter(
+                editor=self.object.editor,
+                date_created__gte=datetime.today() - timedelta(days=90),
+            )
             .exclude(pk=app.pk)
-            .order_by("-id")[:5]
+            .order_by("-id")
         )
+        if len(latest_recent_apps) < 5:
```
suecarmol, line 980: Per @Samwalton9's comment on this PR, we only need to show the applications made in the last 90 days, so we can remove this code (and the test you created for it).

suecarmol (changes requested):
Thank you for your PR! This is looking great, but some changes need to be made in order to merge this 🙂

TWLight/applications/templates/applications/application_evaluation.html
Samwalton9: Maybe `Recent Applications (last 90 days)`?

TWLight/applications/templates/applications/application_evaluation.html
lalit97: removed background color 👍

TWLight/applications/templates/applications/application_evaluation.html
lalit97: removed extra column 👍

TWLight/applications/templates/applications/application_evaluation.html
lalit97: removed borders from table 👍

TWLight/applications/templates/applications/application_evaluation.html
lalit97: looks good to me changing the title to this 🙂

TWLight/applications/views.py
lalit97: Yes :) removed this part and test for it.

suecarmol (approved):
Thank you for the changes you made! This LGTM!

lalit97:
Thank you :)

PR #534, opened 2020-09-24
Update AuthorizedUsers API to only list recently logged in Bundle users

shrey27tri01:

## Description
Added some logic to the [AuthorizedUsers API view](https://github.com/WikipediaLibrary/TWLight/blob/master/TWLight/users/views.py#L675) to filter out the users who are not active (who have not logged in in the last two weeks), that is, list only those users who have logged in the past two weeks. 

## Rationale
Currently, the Authorized Users API view lists all the authorized users. This may work well for approval-based partners since it is application-based. But for the Library Bundle users, where users are granted access automatically, if the user has not been active lately, we do not need to show them as part of the authorized users' list in the Authorized Users API view.

## Phabricator Ticket
[T261219](https://phabricator.wikimedia.org/T261219)

## How Has This Been Tested?
I have not yet tested my changes. One of the existing [tests](https://github.com/WikipediaLibrary/TWLight/blob/0ad014a67a0293fad38a329bda3f0a2a87e1a40a/TWLight/tests.py#L1187) is failing. I wanted to check whether this test is failing due to a limited list of users being returned or there is something wrong in my code. In the first case, then, do I change the existing test, or I add another one?

## Types of changes


shrey27tri01:
In my latest [commit](https://github.com/WikipediaLibrary/TWLight/pull/534/commits/18f574531653ff26d4e6d3efea85e2d5486cffed), I've fixed the issue due to which the test was failing. I've also added another test to check that the users who haven't been active in the last two weeks are not listed in the AuthorizedUsers API view.

TWLight/tests.py
```diff
@@ -33,6 +33,8 @@
     ToURequired,
     EmailRequired,
 )
+import pytz
```
jsnshrmn, line 36: while this works, it's not very idiomatic for Django, which has it's own timezone handling wrappers.
You should use its tooling instead of `pytz`
https://docs.djangoproject.com/en/3.0/topics/i18n/timezones/

TWLight/users/views.py
```diff
@@ -1,6 +1,7 @@
 import datetime
 import json
 import logging
+import pytz
```
jsnshrmn, line 4: thoughts from `TWLight/tests.py` apply here as well.

jsnshrmn (changes requested):
Thanks for your work on this! I did a non-comprehensive read-through and have some feedback inline.

shrey27tri01:
@jsnshrmn Thank you very much for your review! I have made the relevant changes in my latest [commit](https://github.com/WikipediaLibrary/TWLight/pull/534/commits/c24dbaec26785047907b60cdcd3bef8b2eabc5bd) to add support for timezone awareness from `django.utils.timezone` instead of `pytz`. Please have a look at it and suggest further changes, if necessary.

jsnshrmn (changes requested):
It looks like you set all files as executable in d6cbfa0 which is a problem. Please amend this commit so that it only features the file content changes. If you're not sure how to do that let me know, and I'll amend the commit for you.

shrey27tri01:
@jsnshrmn I have reverted the commit [`d6cbfa0`](https://github.com/WikipediaLibrary/TWLight/commit/d6cbfa00f9d6d042bfb1e6ed0493964db5e50bce), so that only the changes in the file contents are featured. Please review the commit and suggest further changes.

TWLight/users/views.py
```diff
@@ -693,6 +694,14 @@ def get(self, request, pk, version, format=None):
...
         # level. This is fine for the use case we built it for (Wikilink tool)
         valid_partner_auths = get_valid_partner_authorizations(pk)
+
+        # List only those users who have been active lately,
+        # that is, do not list the users who had last logged in more than two weeks ago
+        if partner.authorization_method == partner.BUNDLE:
+            valid_partner_auths = valid_partner_auths.filter(
+                user__editor__user__last_login__gt=timezone.now() - timedelta(weeks=2)
+            )
+
```
jsnshrmn, line 704: This code as it stands will fire `get_valid_partner_authorizations(pk)` whether it will be used or not.
Here's a suggested fix that you can commit directly into the PR.

Let's put `get_valid_partner_authorizations` an `else` and update the comments to better reflect the fact that we're dealing with a list of auths, not users.
```suggestion
        # We're ignoring streams here, because the API operates at the partner
        # level. This is fine for the use case we built it for (Wikilink tool)
        valid_partner_auths = None

        # For Bundle partners, get auths for users who logged in within the last 2 weeks.
        if partner.authorization_method == partner.BUNDLE:
            valid_partner_auths = valid_partner_auths.filter(
                user__editor__user__last_login__gt=timezone.now() - timedelta(weeks=2)
            )
        # Otherwise, get all valid auths for this partner.
        else:
            valid_partner_auths = get_valid_partner_authorizations(pk)
```

jsnshrmn (changes requested):
This is looking pretty good! I have one area that I'd like you to change, and I added a suggestion that you can commit from the web interface to get this merge-able.

TWLight/users/views.py
jsnshrmn: also, I just noticed that you were using `user__editor__user__` in your query filter.  I streamlined that to `user__` to eliminate unnecessary db queries.

jsnshrmn:
Hmm, it looks like my guard clause isn't working quite as I expected. It's evening for me here, so I won't take a look until the morning, but know that this is on my radar. Thanks for your work!

shrey27tri01:
@jsnshrmn Sure! I think the test are failing because under the `if` condition we applied, the `filter` can't be applied on a `None` object.

jsnshrmn:
Yeah, this was a boneheaded review from me. Apologies! Your code doesn't even cause additional queryset evaluation (meaning no extra db calls), so really just the comment updates were all that were needed. Going to push the code back to where you had it, but with my comment updates.

jsnshrmn (approved):

PR #540, opened 2020-10-01
Remove the unwnated message of 'Please agree to the term of use'

YashAgrawal9265:

## Description

## Rationale

## Phabricator Ticket

## How Has This Been Tested?

## Types of changes

### DIFF
--- a/TWLight/static/css/local.css
+++ b/TWLight/static/css/local.css
@@ -774,8 +774,3 @@ CLEARING COLUMNS
   color: #fff;
 }
 
-.homeBtn:hover {
-  background-color:#337ab7;
-  border-color:#337ab7;
-  color: #fff;
-}
--- a/TWLight/static/css/local.css
+++ b/TWLight/static/css/local.css
@@ -774,3 +774,8 @@ CLEARING COLUMNS
   color: #fff;
 }
 
+.homeBtn:hover {
+  background-color:#337ab7;
+  border-color:#337ab7;
+  color: #fff;
+}
--- a/TWLight/templates/base.html
+++ b/TWLight/templates/base.html
@@ -137,7 +137,7 @@
         {% if request.GET %}
         <a  class="btn btn-default" href="{% url 'oauth_login' %}?next={{ request.path|urlencode }}&{{ request.GET.urlencode }}">
         {% else %}
-        <a  class="btn btn-default" href="{% url 'oauth_login' %}?next={{ request.path|urlencode }}">
+        <a  class="btn btn-default homeBtn" href="{% url 'oauth_login' %}?next={{ request.path|urlencode }}">
         {% endif %}
           <span class="glyphicon glyphicon-off"></span>
           {% comment %}Translators: Shown in the top bar of almost every page when the current user is not logged in. {% endcomment %}
--- a/TWLight/templates/base.html
+++ b/TWLight/templates/base.html
@@ -137,7 +137,7 @@
         {% if request.GET %}
         <a  class="btn btn-default" href="{% url 'oauth_login' %}?next={{ request.path|urlencode }}&{{ request.GET.urlencode }}">
         {% else %}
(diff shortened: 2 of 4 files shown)


suecarmol:
Hi, @YashAgrawal9265. Thank you for contributing to the project! This message should appear every time you log into the site without having agreed to the terms of use. Did you agree with the terms of use? If not, could you agree to them and then check if the message appears. Let me know if you still see that message after agreeing to the terms of use.

YashAgrawal9265:
Hi, **@suecarmol** . I had also checked that message after agreeing to the term of use. The message did not appears

Samwalton9:
@suecarmol Yash and I chatted about this a little on the Phabricator ticket and I suggested moving it entirely because the yellow message is always present. It felt redundant to also be showing the one-off blue notice when the yellow one is right below it anyway:
image

YashAgrawal9265:
I had made this commit **9dc78c6**. This has pass all the checks .Please review it and suggest if any change needed...

suecarmol:
Thanks for the information. Can I get a link to the Phabricator ticket? I can't seem to find it in our Kanban board

YashAgrawal9265:
I had not metioned this task on the phabricator , so phabricator ticket is not available

Samwalton9:
I misspoke, we spoke via email, not Phabricator :)

TWLight/static/css/local.css
```diff
@@ -774,3 +774,8 @@ CLEARING COLUMNS
   color: #fff;
 }
 
+.homeBtn:hover {
+  background-color:#337ab7;
+  border-color:#337ab7;
+  color: #fff;
+}
```
suecarmol, line 781: I'm not sure about this change here. I believe we should keep the button's default CSS. @Samwalton9 @jsnshrmn what do you think?

This is the current behavior of the default buttons on hover:
Screen Shot 2020-10-06 at 11 52 36

This is the changed behavior of the default buttons on hover:
Screen Shot 2020-10-06 at 11 52 46

If we keep the changes you made, I think the CSS class should be renamed to something different, like `default-btn-override:hover`

suecarmol (commented):
This is looking good, I just need some clarifications on some code

TWLight/static/css/local.css
jsnshrmn: It's not clear to me what problem the css change is solving or how it relates to the pr. I'd exclude the change.

YashAgrawal9265: Should I rename the CSS class right now???

YashAgrawal9265: Actually It's my mistake that I had put this CSS change in this PR. Instead I should file a another PR for this CSS change.
If you don't like this change, I can remove this...

jsnshrmn: Let's back up a step: why did you change the css?

jsnshrmn: Is there a phab issue for the problem that the css change is trying to solve?

YashAgrawal9265: I thought that this change would look nice and more attractive than the default CSS

YashAgrawal9265: No, but I had discuss this change with Sam on email

jsnshrmn: Ah, yeah, I can't really evaluate in that case, but it doesn't seem like an obvious improvement to me since it's a common bootstrap UI metaphor to simply darken the existing color when hovering over an interactive UI element instead of changing colors completely. Will wait to hear back from @Samwalton9 either for evaluation or for more information. I agree with @suecarmol that if we want to accept the change, the class name shouldn't be "homeBtn".

Samwalton9: Agreed that the default behaviour is fine, I don't see a need to add complexity for such a minor change.

YashAgrawal9265:
I had removed the changes of CSS which I had made earlier.

suecarmol (approved):
Thank you for those changes. Moving to merge this PR.

PR #543, opened 2020-10-10
fix the bug in a message which was appearing when user return the aut…

YashAgrawal9265:
…horization

## Description
Users can return an authorization, expiring it and freeing up a space for other users, on their My Library page. Doing so sends them a system message, which currently appears to be broken. 
Before solving the message was looking like as show in the image:

So, I have fix this bug.

## Rationale

## Phabricator Ticket
https://phabricator.wikimedia.org/T264999

## How Has This Been Tested?

## Types of changes


jsnshrmn (approved):
This looks great. Thanks for your contribution!

PR #549, opened 2020-10-15
Add admin validation ensuring that Partners set to the Proxy or Bundle must have a target url

nirmalasainsara:

## Description

## Rationale

## Phabricator Ticket
https://phabricator.wikimedia.org/T236390

## How Has This Been Tested?
yes

## Types of changes

### DIFF
--- a/TWLight/resources/models.py
+++ b/TWLight/resources/models.py
@@ -439,6 +439,13 @@ class Partner(models.Model):
             raise ValidationError(
                 "Partners with automatically sent messages require user instructions to be entered"
             )
+        if self.authorization_method in [self.PROXY, self.BUNDLE] and (
+            not self.specific_stream
+        ):
+            # Validate that target_url should not be empty
+            # when authorization method is PROXY or BUNDLE
+            if not self.target_url:
+                raise ValidationError("Proxy and Bundle partners require a target URL.")
 
     def get_absolute_url(self):
         return reverse_lazy("partners:detail", kwargs={"pk": self.pk})
--- a/TWLight/resources/tests.py
+++ b/TWLight/resources/tests.py
@@ -501,6 +501,52 @@ class PartnerModelTests(TestCase):
             # We're making sure that it fails because of a lack of user instructions, specifically.
             self.assertEqual([error_msg], e.messages)
 
+    def test_for_target_url_not_empty_when_authorization_method_is_proxy_or_bundle(
+        self,
+    ):
+        """
+        When partner’s authorization method is PROXY or BUNDLE
+        then leaving target_url empty should raise a ValidationError
+        """
+        message = "Proxy and Bundle partners require a target URL."
+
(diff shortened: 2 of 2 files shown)


TWLight/resources/models.py
```diff
@@ -439,6 +439,14 @@ def clean(self):
             raise ValidationError(
                 "Partners with automatically sent messages require user instructions to be entered"
             )
+        if (
```
Samwalton9, line 442: This looks good, but could be simplified like:
`if (self.authorization_method in [self.PROXY, self.BUNDLE])`

TWLight/resources/tests.py
```diff
@@ -501,6 +501,35 @@ def test_user_instructions(self):
...
+            True  # We don't want the ValidationError from requested_access_duration
+        )
+
+        self.assertRaises(ValidationError, partner1.clean)
+        try:
+            partner1.save()
+        except ValidationError as e:
+            self.assertEqual([message], e.messages)
+
+        # test for a partner with proxy authorization method
```
Samwalton9, line 525: I think this should be a separate test.

TWLight/resources/tests.py
```diff
@@ -501,6 +501,35 @@ def test_user_instructions(self):
...
+        partner2 = PartnerFactory(authorization_method=Partner.BUNDLE)
+        self.assertRaises(ValidationError, partner2.clean)
+        try:
+            partner2.save()
+        except ValidationError as e:
+            self.assertEqual([message], e.messages)
+
```
Samwalton9, line 532: I think it would also be helpful (in light of a recent bug we found - https://phabricator.wikimedia.org/T265331) to have another test which checks that for a non-PROXY or BUNDLE partner, no error message is raised.

TWLight/resources/models.py
```diff
@@ -439,6 +439,14 @@ def clean(self):
...
+            self.authorization_method == self.PROXY
+            or self.authorization_method == self.BUNDLE
+        ):
```
Samwalton9, line 445: One additional thought I had, which wasn't in the original Phabricator ticket (sorry!), is that we should also check in this if statement that the resource doesn't have streams. If there are streams the streams have `target_url` instead of the resource. This isn't obvious and is a bad design decision from us, but it is the way it is :)

This would just mean adding a `and not self.specific_stream` here. I think it would also be helpful to add another test for this case.

Samwalton9 (changes requested):
This looks great, and definitely works exactly how we'd request! I left a few in-line comments for areas where we could improve a little, and a comment about a specification I had missed in the original Phabricator ticket.

Please let me know if you have any questions about any of my comments :)

nirmalasainsara:
I have made the suggested changes.

jsnshrmn (approved):
This looks good! Thanks for your contribution!

PR #550, opened 2020-10-16
T265146: user_update_eligibility should correctly update editcount for bundle eligibility.

jsnshrmn:
## Description
This limits the scope of editors that may be modified by the command, which should protect us against overrun errors and also reduce the runtime of the command.

## Rationale
Testing revealed that the command incorrectly updates edit counts even if the user has already been updated. This reduces the intended time period used to calculate recent edit counts, effectively making bundle eligibility tougher to keep than it should be.

## Phabricator Ticket
https://phabricator.wikimedia.org/T265146

## How Has This Been Tested?
I finished out the partial timestamp override code so that I could run the command in the test suite against specified dates. I was able to reproduce the issue we saw in production with a failing test, and then get the test passing by updating the command as well as the underlying helper function.

NA

## Types of changes

### DIFF
--- a/TWLight/users/helpers/editor_data.py
+++ b/TWLight/users/helpers/editor_data.py
@@ -12,6 +12,19 @@ logger = logging.getLogger(__name__)
 def editor_global_userinfo(
     wp_username: str, wp_sub: typing.Optional[int], strict: bool
 ):
+    """
+    Public interface for fetching Editor global_userinfo.
+    Parameters
+    ----------
+    wp_username : str
+    wp_sub : int
+    strict : bool
+
+    Returns
+    -------
+    dict
+        Contains the editor's globaluserinfo as returned by the mediawiki api.
+    """
     guiuser = urllib.parse.quote(wp_username)
     # Trying to get global user info with the username
     results = _get_user_info_request("guiuser", guiuser)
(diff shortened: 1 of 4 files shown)


jsnshrmn:
I started to just limit the command to valid editors who haven't been updated in 30 days, but I remembered that we actually do need to check ineligible users too so that we don't let the time delta between counts for recent edits exceed 30 days. So we're swapping out the queryset filter from eligible users to users who haven't been updated in 30 days.

I'll need to add a test for that too and then look more closely at the guard clauses in the `editor_recent_edits` helper.

TWLight/users/helpers/editor_data.py
```diff
@@ -4,14 +4,27 @@
...
 
 
 def editor_global_userinfo(
     wp_username: str, wp_sub: typing.Optional[int], strict: bool
 ):
+    """
+    Public interface for fetching Editor global_userinfo.
+    Parameters
+    ----------
+    wp_username : str
```
suecarmol, line 19: Since you are already adding a docstring, can you include a brief description of each parameter?

TWLight/users/helpers/editor_data.py
```diff
@@ -160,14 +264,16 @@ def editor_recent_edits(
             # This means that eligibility always lasts at least 30 days.
             or (wp_enough_recent_edits and editcount_update_delta.days > 30)
         ):
-            wp_editcount_recent = global_userinfo_editcount - wp_editcount_prev
-            wp_editcount_prev = global_userinfo_editcount
+            # recent edits = global userinfo editcount - stored editcount.
```
suecarmol, line 267: Nitpick: maybe this comment is redundant (at least to me).

suecarmol (dismissed):
This PR is looking great! Adding docstrings to the functions has increased code legibility a lot :) I just have a couple of comments before we can move ahead and merge this.

TWLight/users/helpers/editor_data.py
jsnshrmn: Fair. The comment preceded the code, and was mostly there to help me sort the bug out.
```suggestion
```

TWLight/users/helpers/editor_data.py
jsnshrmn: Done!

jsnshrmn:
Okay, doc/comment changes made. Merging.

PR #555, opened 2020-10-31
Block /admin login page in production & staging

nirmalasainsara:
## Description
No one should allowed to log in via /admin directly, so we should remove that from URLs

## Phabricator Ticket
[//]: # https://phabricator.wikimedia.org/T245268

## How Has This Been Tested?
No

## Types of changes

### DIFF
--- a/TWLight/settings/base.py
+++ b/TWLight/settings/base.py
@@ -44,6 +44,9 @@ TWLIGHT_HOME = os.path.dirname(
 
 TWLIGHT_ENV = os.environ.get("TWLIGHT_ENV")
 
+# To Block/Unblock /admin login page in production & staging
+ADMIN_ENABLED = False
+
 # An atypical way of setting django languages for TranslateWiki integration:
 # https://translatewiki.net/wiki/Thread:Support/_The_following_issue_is_unconfirmed,_still_to_be_investigated._Adding_TheWikipediaLibrary_Card_Platform_TranslateWiki
 
--- a/TWLight/settings/local.py
+++ b/TWLight/settings/local.py
@@ -12,6 +12,9 @@ from .base import *
 
 DEFAULT_FROM_EMAIL = "<twlight.local@localhost.localdomain>"
 
+# To Block/Unblock /admin login page in production & staging
+ADMIN_ENABLED = True
+
 # TEST CONFIGURATION
 # ------------------------------------------------------------------------------
 
--- a/TWLight/tests.py
+++ b/TWLight/tests.py
@@ -1263,3 +1263,38 @@ class AuthorizedUsersAPITestCase(AuthorizationBaseTestCase):
         expected_json = [{"wp_username": self.editor1.wp_username}]
 
         self.assertEqual(response.data, expected_json)
+
+
+class AdminEnableTestCase(TestCase):
+    def test_admin_enabled(self):
+        """
+        Test behaviour when ADMIN_ENABLED set to True
+        """
+
+        # check ADMIN_ENABLED value
+        self.assertEqual(settings.ADMIN_ENABLED, True)
+
+        # check homepage
+        response = self.client.get(reverse("homepage"))
+        self.assertEqual(response.status_code, 200)
+
+        # Check admin page
+        response = self.client.get("/admin/")
+        self.assertRedirects(response, "/admin/login/?next=/admin/")
+        self.assertEqual(response.status_code, 302)
+
+    def test_admin_disabled(self):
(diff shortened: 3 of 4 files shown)


jsnshrmn (changes requested):
The base template for the site expects admin to be available, so you'll have to work on that. When I set `ADMIN_ENABLED = False` and try to browse the site, I get the following error:

```
Request Method: | GET
-- | --
http://localhost/
3.0.11
NoReverseMatch
'admin' is not a registered namespace
/venv/lib/python3.8/site-packages/django/urls/base.py in reverse, line 83
/venv/bin/python
3.8.2
['/venv/bin',  '/app',  '/venv',  '/usr/lib/python3.8',  '/usr/lib/python38.zip',  '/usr/lib/python3.8/lib-dynload',  '/venv/lib/python3.8/site-packages']
```

From a design perspective, `ADMIN_ENABLED = False` should probably be set in `TWLight/settings/base.py` and then changed to true in `TWLight/settings/local.py` only.

jsnshrmn:
Apologies for the slow response. This error came up at any route/url. You can see in the stack trace I posted that I was just at `http://localhost/` Again, I had to flip the setting to False on local to produce the error, but that is what it would be in the live environment. If you are still having trouble reproducing the issue, then I recommend that you create a unit test to check the behavior for the variable set to both True and False. We would need that before we could merge in anyway.

nirmalasainsara:
Hi I have wrote a test and it the changes are working fine on local But I do not found any way to test if "/admin" gives 404 on production.

jsnshrmn:
Hi there,
Thanks for the followup! To check both cases, you'll need to toggle the `ADMIN_ENABLED` setting in your test, and test the behavior for `True` and `False` individually:
https://docs.djangoproject.com/en/3.0/topics/testing/tools/#overriding-settings

nirmalasainsara:
Hi, I wrote another test to check the behaviour when `ADMIN_ENABLED` is set to `False` with help of https://docs.djangoproject.com/en/3.0/topics/testing/tools/#overriding-settings.  

The second test is failing currently maybe because `settings/local.py` is overriding `settings/base.py` at the execution time but I am not sure.

So please have a look into the current tests I have written and if this method does not work I have looked into one more solution in which we can add a custom `middleware` which ensures that only `staff` users can go to the admin URL refer: https://stackoverflow.com/questions/24559531/how-do-i-restrict-access-to-admin-pages-in-django

jsnshrmn:
This is far enough out of date with the base fork that it's a pain for me to run locally. Can you rebase?

nirmalasainsara:
No problem, I have rebased this branch.

jsnshrmn (changes requested):
You still haven't addressed the first issue that I pointed out, which is that you haven't accounted for the optional removal of the admin namespace in the base template. When I change `ADMIN_ENABLED` to `False` in my local environment, this is what I get.
```
Template error:
In template /app/TWLight/templates/base.html, error at line 96
   'admin' is not a registered namespace
   86 :           <ul class="dropdown-menu">
   87 :             <li>
   88 :               <a href="{{ home_url }}">
   89 :                 <i class="fa fa-user-o" aria-hidden="true"></i>
   90 :                 {% comment %}Translators: Shown in the top bar of almost every page inside a dropdown, linking the user to their respective profiles.{% endcomment %}
   91 :                 {% trans "Profile" %}
   92 :               </a>
   93 :             </li>
   94 :             {% if user.is_superuser %}
   95 :               <li>
   96 :                 <a href=" {% url 'admin:index' %} ">
   97 :                   <i class="fa fa-user-plus" aria-hidden="true"></i>
   98 :                   {% comment %}Translators: Shown in the top bar of almost every page inside a dropdown when the current user is a superuser, linking to the administrator interface.{% endcomment %}
   99 :                   {% trans "Admin" %}
   100 :                 </a>
   101 :               </li>
   102 :             {% endif %}
   103 :             <li class="divider"></li>
   104 :             <li>
   105 :               <a href="
   106 :                 {% if user.editor %}
```

jsnshrmn:
...and actually looking at this more closely, you aren't just disabling the admin login route, but all admin routes. We use the admin app in production, so this approach is problematic.

jsnshrmn:
I think this approach is fundamentally a bad fit for our use case. My recommendation would be to just change the nginx config for production and staging to disallow the login routes.

jsnshrmn:
Alternatively, you could just touch the login routes in urls.py
https://stackoverflow.com/a/61946707

PR #566, opened 2020-11-17
Recent edit count isn't updating correctly for some users

jsnshrmn:

## Description
Move editcounts to their own model so that we can manage them effectively.
Update editor model to take advantage of this and to streamline testing.

## Rationale
I had an incorrect mental model of this problem, and we just weren't storing enough data.

## Phabricator Ticket
https://phabricator.wikimedia.org/T265146

## How Has This Been Tested?
I added tests to update editcounts multiple times within an editor's eligibility window.
Tests now call editor methods instead of making a large number of repetitive helper function calls.

N/A

## Types of changes


jsnshrmn:
I can see we actually need a small template change for the editor detail view as well. Hopping on that task now.

jsnshrmn:
okay, that's done. the guard clauses in that template had been implemented a bit strangely, but they look good now, and I manually checked to make sure they were happy.

TWLight/users/migrations/0067_auto_20201120_2002.py
```diff
@@ -0,0 +1,33 @@
+# Generated by Django 3.0.11 on 2020-11-20 20:02
```
suecarmol, line 1: It would be great to change the name of this file to indicate that fields are being removed from the `Editor` model.

TWLight/users/models.py
```diff
@@ -256,6 +223,212 @@ def encode_wp_username(self, username):
...
+        except models.ObjectDoesNotExist:
+            pass
+
+    @property
+    def wp_editcount_updated(self):
+        """
+        Fetches timestamp of latest editcount from EditorLogs related to this editor.
+        Parameters
+        ----------
+        self
```
suecarmol, line 249: Maybe we should remove this, I think docstrings never include `self` params :)

TWLight/users/models.py
```diff
@@ -256,6 +223,212 @@ def encode_wp_username(self, username):
...
+            ).timestamp
+        except models.ObjectDoesNotExist:
+            pass
+
+    def wp_editcount_recent(
+        self,
+        current_datetime: timezone = None,
+    ):
+        """
+        Calculates recent editcount based on EditorLogs related to this editor.
```
suecarmol, line 319: Let's add that the recent edit count is used to determine whether a user may be bundle eligible or not.

suecarmol (commented):
This is a great PR! Thank you for refactoring the recent edit counts. I have just a few comments that are a bit nit-picky. Other than that, this looks good to me.

TWLight/users/models.py
jsnshrmn: That's weird, as I just used the auto-docstring function in pycharm. Those were there in some places, but not others. I had a bunch of editor updates that I applied midway through, and it doesn't seem to be doing it now. Definitely doesn't make any sense. Gone!

TWLight/users/models.py
jsnshrmn: done!

TWLight/users/migrations/0067_auto_20201120_2002.py
jsnshrmn: Fair. Done!

PR #571, opened 2020-11-24
Withdraw button will appear only to the status which are PENDING,APPR…

YashAgrawal9265:

## Description
Withdraw was appearing to all the status except invalid. I make a change in it. Now it will appear to only  PENDING, APPROVED and  UNDER DISCUSSION status.

## Rationale

## Phabricator Ticket
https://phabricator.wikimedia.org/T268449

## How Has This Been Tested?

## Types of changes
- [x ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/TWLight/users/templates/users/my_applications.html
+++ b/TWLight/users/templates/users/my_applications.html
@@ -58,7 +58,7 @@
           {% trans 'Not yet reviewed.' %}
         {% endif %}
         <br />
-        {% if app.status != app.INVALID %}
+        {% if app.status == app.APPROVED or app.status == app.PENDING or app.status == app.QUESTION %}
           <form action="{% url 'users:withdraw' user.editor.pk app.pk%}" class="m-2" method="post">
           {% csrf_token %}
           <input type="submit" value="Withdraw" class="btn btn-danger btn-sm">
--- a/TWLight/users/tests.py
+++ b/TWLight/users/tests.py
@@ -312,8 +312,18 @@ class ViewsTestCase(TestCase):
         # expected string is too fragile.
 
     def test_withdraw_application(self):
-        app = ApplicationFactory(
+        app1 = ApplicationFactory(
             status=Application.PENDING,
+            partner=PartnerFactory(authorization_method=Partner.BUNDLE),
+            editor=self.user_editor.editor,
+        )
+        app2 = ApplicationFactory(
+            status=Application.APPROVED,
+            partner=PartnerFactory(authorization_method=Partner.BUNDLE),
+            editor=self.user_editor.editor,
+        )
+        app3 = ApplicationFactory(
+            status=Application.QUESTION,
             partner=PartnerFactory(authorization_method=Partner.BUNDLE),
             editor=self.user_editor.editor,
         )
@@ -326,7 +336,9 @@ class ViewsTestCase(TestCase):
             request, pk=self.editor1.pk, id=app.pk
         )
         app.refresh_from_db()
-        self.assertEqual(app.status, Application.INVALID)
+        self.assertEqual(app1.status, Application.INVALID)
+        self.assertEqual(app2.status, Application.INVALID)
+        self.assertEqual(app3.status, Application.INVALID)
 
     def test_my_library_page_has_authorizations(self):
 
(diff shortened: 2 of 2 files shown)


lalit97:
Minor suggestion, instead of 
`{% if app.status == app.APPROVED or app.status == app.PENDING or app.status == app.QUESTION %}`
We can write the condition something like 
`{% if app.status in [app.APPROVED, app.PENDING, app.QUESTION] %}`

YashAgrawal9265:
@lalit97 Thank You! for your suggestion but it is showing template syntax error.

YashAgrawal9265:
This syntax might work on JINJA TEMPLATE but not on Django templates.

lalit97:
> @lalit97 Thank You! for your suggestion but it is showing template syntax error.

Oh my bad, I suggested this after reading https://docs.djangoproject.com/en/1.11/ref/templates/builtins/#in-operator but looks like it does not work for lists :(

jsnshrmn (changes requested):
Hi there,

Thanks for your work on this, and apologies for the long wait for review. Since the underlying issue that precipitated this change was that we should not be sent applications to be withdrawn, it warrants adding a test to verify that is so. You could add it right after the existing withdraw test:
https://github.com/WikipediaLibrary/TWLight/blob/cdc42e01085516978d1a7842b96352d5bf94a1ae/TWLight/users/tests.py#L317

YashAgrawal9265:
I have made some changes in the test case.
Please have a look on it.

jsnshrmn:
The we need here is one for an app with the status of `SENT`. We want to test to make sure that those may not be withdrawn.

YashAgrawal9265:
Instead of 4 apps, should I add only 1 app with status as Sent?

YashAgrawal9265:
test for app with status as **sent** cannot be made because when application status is **sent**, in that case withdraw button is not available.

jsnshrmn:
I'm asking for a test that validates that the button is not there in this situation.

YashAgrawal9265:
how I can check that button is available or not in the test? Can you give me some brief idea?

jsnshrmn:
Oh yes, I apologise for omitting an example. The way we've handled this elsewhere in the project is to render the view and then do an `assertNotIn`
Example:
https://github.com/WikipediaLibrary/TWLight/blob/350e5f1ceaa9e4355605aa3e267e4c0dcf6884fc/TWLight/applications/tests.py#L3189

YashAgrawal9265:
I have pushed one commit. Please tell me that what mistake I am making in testcase. It is giving me error.

jsnshrmn:
Two thoughts here:
1) You have replaced the original test case, meaning that you have removed the test that ensures that the withdraw button works. You should be adding a new test rather than replacing the existing one.
1) The issue here is that you've created an application that was sent by no one. If you look at the validation error, you can see that the resulting authorization is failing out for this reason. Normally the authorizer is set to the coordinator who sends the application. You need to set `sent_by` in the application factory to avoid this.

YashAgrawal9265:
I have add one more commit, Please check it.
As you said I have add a different testcase for checking withdraw button for **sent** status.
But now I am getting this error.



Might this error is due to **assertNotIn** . Can you tell me what should I write in **assertNotIn**.

jsnshrmn:
I'm going to encourage you to do some of the legwork on figuring this out. That error is indicating that you are receiving a redirect rather than renderable output. You could try to print some information to your console to find more information about the redirect (status code, in body messages, etc), you could look at the source for the view and the route to see if you can figure out why you are receiving a redirect. If you understand why the redirect is happening, then you can decide if you should follow it, or if you need to do something else in the request setup to prevent it.

YashAgrawal9265:
Can I use logger in the test.py

YashAgrawal9265:
Can I use logger in the test.py?

jsnshrmn:
I usually just throw in a `print()` statement for quick debugging of tests.

YashAgrawal9265:
I found that first I was using **WithdrawApplicationView**  due to which I always get redirect response. So now I have add **ListApplicationView**.
Please check. If I am wrong again, the please correct me.

jsnshrmn (approved):

jsnshrmn (approved):
This looks great! Thanks for your work!

PR #573, opened 2020-11-30
Remove activity feed

YashAgrawal9265:

## Description
In an attempt to emulate the transparency of Wikipedia, we implemented an activity feed showing recent applications, approvals, and signups (https://wikipedialibrary.wmflabs.org/activity/). We've never seen much interest in this page, and in a world where we have the Library Bundle makes less sense as a view of how people are using the tool.

In an effort to simplify the tool and remove bloat, we should simply remove the activity page and all associated views & unused code. This includes the application option to hide the application from the activity feed.

So I have remove the activity feed.

## Rationale

## Phabricator Ticket
https://phabricator.wikimedia.org/T213202

## How Has This Been Tested?

## Types of changes


jsnshrmn (approved):
Thanks for your contribution! I was initially concerned when I saw that there were no changes to the test suite, but when I looked closely, I found that we had no tests for the activity feed. How embarrassing for us, huh? In addition to helping us de-clutter our site, this change will improve our test coverage numbers by removing untested code. 🎉

PR #577, opened 2020-12-07
Fix server error when coordinators view their own application

nirmalasainsara:
## Description
When coordinators view their own application to a Partner which
has no coordinator then we need to return None

## Phabricator Ticket
https://phabricator.wikimedia.org/T262248

## How Has This Been Tested?
I have not tested it.

## Types of changes


jsnshrmn (approved):
Normally, I'd ask for a test, but this is a small change that fixes a server error and is something I'd probably catch in a hotfix if I were working on it. Thanks!

PR #581, opened 2020-12-11
Don't send two separate emails when access code applications are appr…

YashAgrawal9265:

## Description
We send users an email when their account is approved. We also send them an email when their application is marked as 'Sent'. In the case of access code applications, these two actions can happen very soon after each other - we already have the access codes so there's no need to send the user's details over to some external partner. We can approve their application, and then head over to the Send interface to assign access codes. Ideally, therefore, we should only send the user one email in the access codes case - when their application is marked Sent with the corresponding access code.

## Rationale

## Phabricator Ticket
https://phabricator.wikimedia.org/T262521

## How Has This Been Tested?

## Types of changes

### DIFF
--- a/TWLight/emails/tasks.py
+++ b/TWLight/emails/tasks.py
@@ -61,10 +61,6 @@ class CommentNotificationEmailOthers(template_mail.TemplateMail):
     name = "comment_notification_others"
 
 
-class ApprovalNotification(template_mail.TemplateMail):
-    name = "approval_notification"
-
-
 class WaitlistNotification(template_mail.TemplateMail):
     name = "waitlist_notification"
 
@@ -316,32 +312,6 @@ def send_comment_notification_emails(sender, **kwargs):
             )
 
 
-def send_approval_notification_email(instance):
-    base_url = get_current_site(None).domain
-    path = reverse_lazy("users:my_library")
-    link = "https://{base}{path}".format(base=base_url, path=path)
-    email = ApprovalNotification()
-
-    # If, for some reason, we're trying to send an email to a user
-    # who deleted their account, stop doing that.
-    if instance.editor:
-        email.send(
-            instance.user.email,
-            {
-                "user": instance.user.editor.wp_username,
-                "lang": instance.user.userprofile.lang,
-                "partner": instance.partner,
-                "link": link,
-                "user_instructions": instance.get_user_instructions(),
-            },
-        )
-    else:
-        logger.error(
-            "Tried to send an email to an editor that doesn't "
-            "exist, perhaps because their account is deleted."
-        )
-
-
 def send_waitlist_notification_email(instance):
     base_url = get_current_site(None).domain
     path = reverse_lazy("partners:filter")
@@ -418,7 +388,6 @@ def update_app_status_on_save(sender, instance, **kwargs):
     """
     # Maps status indicators to the correct email handling function.
     handlers = {
-        Application.APPROVED: send_approval_notification_email,
         Application.NOT_APPROVED: send_rejection_notification_email,
         # We can't use Partner.WAITLIST as the key, because that's actually an
         # integer, and it happens to be the same as Application.APPROVED. If
@@ -462,6 +431,9 @@ def update_app_status_on_save(sender, instance, **kwargs):
 
 @receiver(pre_save, sender=AccessCode)
 def send_authorization_emails(sender, instance, **kwargs):
+    base_url = get_current_site(None).domain
+    path = reverse_lazy("users:my_library")
+    link = "https://{base}{path}".format(base=base_url, path=path)
     """
     When access code objects are updated, check if they should trigger
     an email to a user.
@@ -483,6 +455,9 @@ def send_authorization_emails(sender, instance, **kwargs):
                     "partner": instance.partner,
                     "access_code": instance.code,
                     "user_instructions": instance.partner.user_instructions,
+                    "link": link,
+                    # "user": instance.au.editor.wp_username,
+
                 },
             )
             email.send()
--- a/TWLight/emails/tasks.py
+++ b/TWLight/emails/tasks.py
@@ -61,6 +61,10 @@ class CommentNotificationEmailOthers(template_mail.TemplateMail):
     name = "comment_notification_others"
 
 
+class ApprovalNotification(template_mail.TemplateMail):
+    name = "approval_notification"
+
+
 class WaitlistNotification(template_mail.TemplateMail):
     name = "waitlist_notification"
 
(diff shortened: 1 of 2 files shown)


TWLight/emails/tasks.py
```diff
@@ -61,10 +61,6 @@ class CommentNotificationEmailOthers(template_mail.TemplateMail):
     name = "comment_notification_others"
 
 
-class ApprovalNotification(template_mail.TemplateMail):
```
Samwalton9, line 64: This email is still used in cases where the publisher isn't set to the access codes authorization method - we shouldn't be deleting it outright, just handling the access codes case better.

TWLight/emails/tasks.py
```diff
@@ -316,32 +312,6 @@ def send_comment_notification_emails(sender, **kwargs):
             )
 
 
-def send_approval_notification_email(instance):
```
Samwalton9, line 319: Likewise this is still needed, we shouldn't delete it.

YashAgrawal9265:
Hi Sam,
I have done necessary changes.
Now the approved email will only sent to application which have authorization method other than access code.

TWLight/emails/templates/emails/access_code_email-body-text.html
```diff
@@ -1,9 +1,17 @@
...
 {% comment %}Translators: This email is sent to users with their access code. Don't translate Jinja variables in curly braces like {{ user }} or {{ partner }}. Translate Wikipedia Library in the same way as the global branch is named (click through from https://meta.wikimedia.org/wiki/The_Wikipedia_Library).{% endcomment %}
 {% blocktrans trimmed %}
+Dear {{ editor_wp_username }},
+
 Your approved application to {{ partner }} has now been finalised. Your access code or login details can be found below:
 
-{{ access_code }}
+{{ access_code }}.
+
+User instructions are {{user_instructions}}.
```
jsnshrmn, line 10: @Samwalton9 Do you think this could just be `{{user_instructions}}`? `User instructions are` is going to an awkward fit with some of the `{{user_instructions}}` and this wrapping text would also get sent to translatewiki, adding another layer of potential odd grammar and extra work.
@YashAgrawal9265, below is an example of how some of our `user_instructions` are written.
```
Log in with the credentials you'll receive shortly at https://example.com/. Upon logging in for the first time, please take an example action. You are welcome to take another example action.
```

TWLight/emails/tasks.py
```diff
@@ -324,7 +324,7 @@ def send_approval_notification_email(instance):
 
     # If, for some reason, we're trying to send an email to a user
     # who deleted their account, stop doing that.
-    if instance.editor:
+    if instance.editor and instance.partner.authorization_method!=1:
```
jsnshrmn, line 327: This will throw an error when we approve applications for partners with an authorization method other than `CODES`. That seems like a problem to me. Also, when you refer to the `authorization_method` please use the choice label instead of the number for legibility. You can just search the codebase for `Partner.CODES` to see plenty of examples.

jsnshrmn (changes requested):
Thanks for your work on this, and apologies for the long wait for review.
Please check the inline comments for questions or things that need addressed.
Also, please note that this needs unit tests for the new behavior to be merge-able.

TWLight/emails/tasks.py
jsnshrmn: Yes, that was mistaken feedback based on a readthrough. It looks as if the error would be logged if the auth type *is* `codes`. You've achieved the goal of not sending a email here for `codes`, but you are doing it in a guard clause that is designed to log errors and send the dev team email alerts. If you check the error output in docker, you'll see that where the current codebase sends the extra email to a user, your code is now logging an error incorrectly.

YashAgrawal9265:
Before making changes in the codebase. We get this mail in the docker.

After making changes, We get this mail in the docker.



What is the error in this?

jsnshrmn:
I would expect this:
https://github.com/WikipediaLibrary/TWLight/blob/dfdeeb22d8e5502fc4f4efdaa835d41ab640333d/TWLight/emails/tasks.py#L338-L342
to come into play

jsnshrmn:
check again with an email type auth

YashAgrawal9265:
I cannot see this logger error before and after making changes in the codebase?
How can I see this?
I am using email with authorization method `codes` and `email`.

jsnshrmn:
I'm going to try to be really explicit here: there are problems with this block of code.
1) it's hard to read for code style reasons that I've already asked you to address. That makes it harder to review.
2) You've co-opted a guard clause that's designed to log an error under certain circumstances, but you haven't updated the comments to reflect the aims of the code block. Again, that makes it harder to review.
3) You don't have any tests. That is how you prove to me that your code works, not screenshots. Again, that makes it harder to review.

Maybe there is some other unexpected behavior that is precluding the error being logged, or maybe I'm just having trouble understanding the code because of the aforementioned issues. This PR cannot get any more attention until these concerns have been addressed.

YashAgrawal9265:
What type of screenshot should I upload?

YashAgrawal9265:
Can you just tell me what this function do?
`self.assertTrue(mock_email.called)`
and 
`self.assertEqual(mock_email.call_count, 1)`

jsnshrmn:
> What type of screenshot should I upload?

You can just mark this as not applicable since there's no visual change here.

YashAgrawal9265:
See this I have generated the test case:

**@patch("TWLight.emails.tasks.send_approval_notification_email")
    def test_approval_calls_email_function(self, mock_email):
        app = ApplicationFactory(status=Application.PENDING)
        app.status = Application.SENT
        app.save()
        self.assertTrue(mock_email.called)**

I have to check this testCase for only email with authorization method as `CODES`. Can you tell me how can I do that?

jsnshrmn:
> I have to check this testCase for only email with authorization method as `CODES`. Can you tell me how can I do that?

We deal with this issue indirectly a number of times in our application and email tests, one method we use is counting the number of items in the outbox.

YashAgrawal9265:
See this :
**@patch("TWLight.emails.tasks.send_approval_notification_email")
    def test_approval_calls_email_function(self, mock_email):
        orig_outbox = len(mail.outbox)
        app = ApplicationFactory(status=Application.PENDING)
        app.status = Application.SENT
        app.save()
        self.assertEqual(len(mail.outbox), orig_outbox+1)*

Do I have make any mistake in this testCase?

YashAgrawal9265:
django.core.exceptions.ValidationError: {'authorizer': ['This field cannot be blank.'], '__all__': ['Authorization authorizer must be a coordinator or staff.']}

jsnshrmn:
> **django.core.exceptions.ValidationError: {'authorizer': ['This field cannot be blank.'], '**all**': ['Authorization authorizer must be a coordinator or staff.']}**
> Can you tell me why I am getting this error when I am running the testcase. and it is also showing the error in
> app.save().

I believe you will find the answer to this question if you look at the other tests that deal with sent applications.

YashAgrawal9265:
Before running any test case do I need to run any command which is used to refresh the database

jsnshrmn:
> Before running any test case do I need to run any command which is used to refresh the database

No, but related to that authorizer bug: I noticed that your fork is fairly out of date, and we did squash a bug related to authorization validation not that long ago. A rebase might be a good idea.

YashAgrawal9265:
I have rebase my branch. I have run testCase for the email without generating any new testCase by me, then also it is showing 
me error.

YashAgrawal9265:
After rebasing I am getting this error:

jsnshrmn:
It looks like you need to run migrations

YashAgrawal9265:
I have run migration script. Now everything is running on my localhost. But when I am running any testCases not only for emails
but for any other folder like `applications` or `resources`. I am getting this error.


Please can you help me out?

jsnshrmn:
Closed due to inactivity.

PR #584, opened 2020-12-21
'Sent' in the cases where the application is to a publisher with the …

YashAgrawal9265:
…EMAIL authorization methods, and 'Finalized' when they have the CODES, PROXY, or LINK methods

## Description
As noted a number of times  'sent' (or 'sent to partner') is becoming a misnomer for the final status of an application on the Library Card platform. Of all our distribution workflows, it really only applies to the EMAIL distribution method. For access codes, proxy, and links, we don't send anything to the partner, and in fact approving an application marks it as 'sent' immediately.

We should rename all strings and variables from 'sent' to 'finalised'. In the EMAIL distribution method case, we could still show users the 'sent to partner' text, and show 'finalized' for other authorization methods.
## Rationale

## Phabricator Ticket
https://phabricator.wikimedia.org/T237517

## How Has This Been Tested?

## Types of changes
- [x ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/TWLight/applications/templates/applications/application_list_include.html
+++ b/TWLight/applications/templates/applications/application_list_include.html
@@ -9,9 +9,15 @@
 {% for app in object_list %}
   <div class="row">
     <div class="col-xs-4 col-sm-2">
+    {% if app.partner.authorization_method == 0 %}
       <span class="label label{{ app.get_bootstrap_class }} pull-right">
         {{ app.get_status_display }}
       </span>
+    {% else %}
+      <span class="label label{{ app.get_bootstrap_class }} pull-right">
+        Finalized
+      </span>
+    {% endif %}
     </div>    
     <div class="col-xs-8 col-sm-6">
       <h4>


TWLight/applications/templates/applications/application_list_include.html
```diff
@@ -9,9 +9,15 @@
 {% for app in object_list %}
   <div class="row">
     <div class="col-xs-4 col-sm-2">
+    {% if app.partner.authorization_method == 0 %}
```
jsnshrmn, line 12: when you refer to the `authorization_method` please use the choice label instead of the number for legibility. You can search for `partner.authorization_method` in html files for examples.

jsnshrmn (changes requested):
Thanks for your work on this, and apologies for the long wait for review.
Please check the inline comments for questions or things that need addressed.

There are other templates that use `get_status_display`, but you haven't updated them with this same logic. Instead of baking this into each template that displays the application status, I'd recommend putting the logic into the view layer and letting the templates stay as simple as possible.

There should be unit tests that ensure that the correct label is rendered into the views where expected.

YashAgrawal9265:
Can you give me some hints, that how should I implement this in the views.py file?
In which class I should make change `_BaseListApplicationView` or `ListApplicationsView`?
Should I make any change in the `baseqs`?

jsnshrmn:
having come back around to take a closer look at this, I can see that this is using a built-in django model method. As a starting place I'd recommend:
updating the status text for `SENT` like so:
```
    # Translators: This is the status of an application that has been finalised for a partner that uses an authorization method other than email.
    (SENT, _("Finalised")),
```
and overriding the method in the app model, like so:
```
def get_status_display(self):
    if self.partner.authorization_method == Partner.EMAIL:
        # Translators: This is the status of an application that has been finalised for a partner using email-based authorization.
        value = _("Sent to partner")
    else:
        field_object = self._meta.get_field('status')
        value = self._get_FIELD_display(field_object)

    return value
```
You could revert your current template change, run migrations, and see how far that gets you. You could then see if there are any views or templates that need updated to use the new label. If they are manually setting status text in the template, you could switch them over to using this method too.

jsnshrmn:
Closed due to inactivity.

PR #585, opened 2020-12-23
Disallow a user from applying to the same partner more than once

nirmalasainsara:
## Description
If a user attempts to load the application request view, but already has an application with the status PENDING, APPROVED, or QUESTION, then we should redirect them to that application. This should only happen for resources with no collections, and which don't have specific_title = True.

## Phabricator Ticket
https://phabricator.wikimedia.org/T239504

## How Has This Been Tested?
yes

## Types of changes


jsnshrmn (approved):
Thanks for your work on this, and apologies for the long wait for review. This looks great!

nirmalasainsara:
No problem, Thank you!

PR #592, opened 2021-01-21
Handle OAuth login errors better

jsnshrmn:

## Description
Moves a bunch of exceptions over to suspicious operation exceptions and updates the 400 error handler to use a slightly modified view that allows the django messages framework to function for 400 errors. Adds a "please try again" message when users encounter the handshaker problem we've been seeing in the logs.

## Rationale
Our unhandled oauth exceptions are unhelpful to users and are blowing up our inboxes with error messages.

## Phabricator Ticket
https://phabricator.wikimedia.org/T255431

## How Has This Been Tested?
I manually raised various exceptions in different parts of the oauth code block to verify the behavior. Oauth exchange is a big black hole in our test suite, and I wasn't about to open that can of worms to fix this annoyance.

This turns out to be a hassle to get good screenshots of for a variety of reasons. Know that the current result is a 500 error, and the PR gives a 400 bad request error with a call to action.

## Types of changes

### DIFF
--- a/TWLight/templates/base.html
+++ b/TWLight/templates/base.html
@@ -180,7 +180,7 @@
         <div class="alert alert-{{ message.tags }}">{{ message|safe }}</div>
       {% endfor %}
     {% endif %}
-    {% if not user.email and not user.is_anonymous %}
+    {% if user.is_authenticated and not user.email %}
       <div class="alert alert-warning">
         {% url 'users:email_change' as email_url %}
         {% url 'contact' as contact_us_url %}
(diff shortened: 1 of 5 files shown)


suecarmol (commented):
This looks good, although I would like to test it in staging before we merge this into production. Could you please merge this to the staging branch so we can test this there?

jsnshrmn:
This is already on staging, but I haven't been able to reproduce an oauth error there. I tried setting the clock back (which is what i did locally), but that doesn't stick on mw managed hosts.

suecarmol (approved):
Got it! Since it doesn't look like there were any other errors on staging, I'll go ahead and merge this! Thanks for a great PR that will keep our inboxes less full :)

PR #598, opened 2021-01-28
block admin login in production

nirmalasainsara:

## Description
 No one should be allowed to log in via /admin directly, so we should remove that from URLs

## Phabricator Ticket
  https://phabricator.wikimedia.org/T245268

## How Has This Been Tested?
 No

## Types of changes

### DIFF
--- a/conf/production.nginx.conf
+++ b/conf/production.nginx.conf
@@ -65,4 +65,7 @@ server {
         root /app/500/;
     }
 
+    location /admin/login {
+      return 403;
+    }
 }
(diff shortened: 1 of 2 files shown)


nirmalasainsara:
I have made changes in Nginx files of staging and production as you suggested in https://github.com/WikipediaLibrary/TWLight/pull/555. But haven't tested the changes if it is working fine since there was no way to do so.

jsnshrmn (approved):
I tested this manually by applying this configuration block to `conf/local.nginx.conf` and it worked as expected. We don't get the styled 403 page provided by the django app, but that's not a problem considering that nobody should be navigating to this route.

nirmalasainsara:
cool, thanks 🙂

PR #603, opened 2021-02-01
WIP Create management command for project page announcement

jsnshrmn:

## Description
Creates a management command, email task, and email templates for project page announcement.

## Rationale
We want to invite active editors to participate in our upcoming projects for the next two quarters.

## Phabricator Ticket
https://phabricator.wikimedia.org/T272367

## How Has This Been Tested?
I added some basic unit tests.
I also manually ran the command in my local environment, and it generated emails.

N/A

## Types of changes


Samwalton9:
I wonder if we shouldn't bother tagging the emails for translation. I'd like to send this out ASAP, before any translations are going to get merged, and then we're left with a bunch of text that will be a waste of time for folks to translate.

jsnshrmn:
> I wonder if we shouldn't bother tagging the emails for translation. I'd like to send this out ASAP, before any translations are going to get merged, and then we're left with a bunch of text that will be a waste of time for folks to translate.

That seems reasonable to me. I'll make that change (and add comments so that we know they aren't marked for translation on purpose) after I rebase later today (or early tomorrow)

PR #609, opened 2021-02-13
Fix capitalization in an English message

amire80:

## Description

Replace "Cannot" with "cannot" in the middle of a sentence.

I do a lot of changes of this kind in MediaWiki core and extensions, and I've never done it with .po files, and it's quite different. It appeared multiple times, and I replaced everywhere. I'm not sure if it's totally right to do it like this, so please feel free to correct me if I'm wrong or to discard the change entirely.

## Rationale

It looks like "Cannot" doesn't need to be capitalized in the middle of a sentence.

## Types of changes

### DIFF
--- a/TWLight/applications/views.py
+++ b/TWLight/applications/views.py
@@ -1076,7 +1076,7 @@ class EvaluateApplicationView(
                 self.request,
                 messages.ERROR,
                 # Translators: this message is shown to coordinators who attempt to change an application's Status from INVALID to any other Status.
-                _("Status of INVALID applications Cannot be changed."),
+                _("Status of INVALID applications cannot be changed."),
             )
             return HttpResponseRedirect(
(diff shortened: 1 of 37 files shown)


TWLight/applications/views.py
```diff
@@ -1076,7 +1076,7 @@ def post(self, request, *args, **kwargs):
                 self.request,
                 messages.ERROR,
                 # Translators: this message is shown to coordinators who attempt to change an application's Status from INVALID to any other Status.
-                _("Status of INVALID applications Cannot be changed."),
+                _("Status of INVALID applications cannot be changed."),
```
Samwalton9, line 1079: This is actually the only change you need to make - the .po files are all generated automatically when the string update is deployed :)

Samwalton9: (Which is why there are now merge conflicts)

amire80:
Thanks! I'll redo it.

PR #613, opened 2021-02-16
Remove trailing spaces in about.html

amire80:
This is preparation for another change in a message.


suecarmol (approved):
Thanks for your contribution!

PR #614, opened 2021-02-16
Remove a space from inside a link in a message

amire80:

## Description

Remove space in a message inside an HTML element.

## Rationale

The current code adds an unnecessary space to an already-non-trivial translatable message. Without the space it will be easier to translate.


suecarmol (approved):
Thanks for making our content easier to translate!

PR #621, opened 2021-02-19
Replace "stream" with "collection" in messages

amire80:

## Description

Replace "stream" with "collection" in messages

Also remove some '(s)' in parentheses in the same messages: it's not so necessary in English, and it's a bit confusing for translators.

## Rationale

Make the text more consistent and easier to translate.

## Phabricator Ticket
[//]: # https://phabricator.wikimedia.org/T275040

### DIFF
--- a/TWLight/applications/helpers.py
+++ b/TWLight/applications/helpers.py
@@ -123,7 +123,7 @@ SEND_DATA_FIELD_LABELS = {
     # Translators: When sending application data to partners, this is the text labelling a user's affiliation
     AFFILIATION: _("Affiliation"),
     # Translators: When sending application data to partners, this is the text labelling the stream/collection a user requested
-    SPECIFIC_STREAM: _("Stream requested"),
+    SPECIFIC_STREAM: _("Collection requested"),
     # Translators: When sending application data to partners, this is the text labelling the specific title (e.g. a particular book) a user requested
     SPECIFIC_TITLE: _("Title requested"),
     # Translators: When sending application data to partners, this is the text labelling whether a user agreed with the partner's Terms of Use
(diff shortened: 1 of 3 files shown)


jsnshrmn (changes requested):
Hi there, and thanks for your work on this!
Unfortunately, you haven't caught all instances of `stream` that editors and coordinators see.
To catch them all, you'll need to search for the word stream across all project files and determine if each occurrence is something a person would see. For example, in the resources models:
https://github.com/WikipediaLibrary/TWLight/blob/master/TWLight/resources/models.py
there are several references to streams that coordinators would see (the help message, the validation error messages, etc). This was just a quick, non-exhaustive check, so you'll need to look for other examples.

amire80:
Thanks for the review, @jsnshrmn.

My concern is mainly with the volunteer translators' comfort. I went over the messages that are _translatable_ in translatewiki, and I'm pretty sure that I caught them all, but please correct me if I'm wrong.

The strings in https://github.com/WikipediaLibrary/TWLight/blob/master/TWLight/resources/models.py appear to be internal and not translatable. Perhaps they are shown to developers or to API users, but not to end users. I'm not even sure they should be changed—maybe the word "stream" is correct for describing internal objects.

jsnshrmn:
Thank you for sharing your motivation for working on this issue. Please understand that when you are working on issues that we've identified, the scope of the issue may not perfectly align with your motivation for working on it. In this case, we've found that the word `stream` has been a point of confusion across the board, not just in public facing, translatable strings. It's undesirable for our internal team to be presented with different language for collections than what editors are presented with. Please include the requested changes. If there is some part of it there is giving you trouble, let us know and we'll try to help; if you are uninterested in working on it now that we've had this discussion, you are free to close this pr and unassign yourself from the task.
We really do appreciate the work you've done on this so far, though, and hope you can push through to completion.
Thanks!

amire80:
I can try doing, but can this be done in several steps, with this patch being the first one? I am more certain about it, and it has immediate benefit to translators. I'll try changing the rest of the code later.

If you really want to do it in one big patch then I guess it can be closed.

jsnshrmn:
That is a reasonable request. The changes you have here are good, they just aren't enough to close out the task. We usually try to map PRs to tasks 1:1, but we can merge this in and make a note in phab that there's still work to do here.

jsnshrmn (approved):
This doesn't completely fix the issue, but it does reduce it. Thanks!

amire80:
Thanks ❤️

PR #630, opened 2021-02-24
Remove a double space from a message

amire80:

## Description

## Rationale

Remove a double space from a message. This makes a translatable message more consistent with other messages.


jsnshrmn (approved):
Good catch. Thanks for your work!

PR #642, opened 2021-03-09
Fix bug where user is not eligible for 6+ months criteria despite account creation in 2006

suecarmol:
## Description
This PR now checks whether a user's account is already over 6 months old before trying to update it since this criterion should only be checked if it's false (once it's true, it should remain true).

## Rationale
We believe the reason a user with an account created in 2006 is not eligible was because of something going wrong in the login process that made the `wp_account_old_enough`  attribute False. Now, if `wp_account_old_enough` is True, it will never be evaluated again. This attribute will also be checked in the `user_update_eligibility` command that runs every day when it is False.

## Phabricator Ticket
[T274625](https://phabricator.wikimedia.org/T274625)

## How Has This Been Tested?
Tested using previous tests to make sure nothing got broken.

N/A

## Types of changes


jsnshrmn (approved):
Looks good to me!

PR #646, opened 2021-03-11
Remove one-time EZProxy announcement email

sahilgrewal8072:
Out of the 8 files, 4 are modified and 4 are deleted succesfully.

Bug:T275041


TWLight/users/models.py
```diff
@@ -121,8 +121,6 @@ class Meta:
     )
     # Temporary field to track sending of project page 2021 email to prevent duplication in case of error.
     project_page_2021_notification_sent = models.BooleanField(default=False)
-    # Temporary field to track sending of proxy launch email to prevent duplication in case of error.
-    proxy_notification_sent = models.BooleanField(default=False)
```
suecarmol, line 125: Since you deleted a field on a model, you need to run `python manage.py makemigrations` and `python manage.py migrate` to update the database.

suecarmol (changes requested):
Thank you for your contribution! There is one important change you need to make in this PR before we can merge this.

TWLight/users/models.py
Samwalton9: Our [documentation](https://github.com/WikipediaLibrary/TWLight/wiki/Branches) says migration files aren't required - does that need updating?

sahilgrewal8072: @suecarmol Done !

suecarmol (approved):
Since we don't need a migration file (it will be created by CI/CD), I will go ahead and merge this. Thanks for your contribution!

PR #648, opened 2021-03-13
If an authorization is to an EMAIL, CODES, or LINK authorization method partner, the Renew/Extend button should always be accessible

suecarmol:
## Description
Changed the behavior of the `apply` and `has_auths` flags so that the Apply button is enabled for partners with certain authorization methods.

Fixed the resource tile partial view to show the `Renew`, `Extend` or `View application` buttons on individual access resources.

## Rationale
[//]: # (Why is this change required? What problem does it solve?)
Before, it was impossible to ask for renewal if the editor already had authorization (even if It had already expired).

## Phabricator Ticket
[//]: # (Link to the Phabricator ticket)
[T271088](https://phabricator.wikimedia.org/T271088)
[T270649](https://phabricator.wikimedia.org/T270649)

## How Has This Been Tested?
[//]: # (- Did you add tests to your changes? Did you modify tests to accommodate your changes?)
[//]: # (- Can this change be tested manually? How?)
1. Create an authorization from a partner with  EMAIL, CODES, or LINK authorization method.
2. Create another authorization from a partner with  EMAIL, CODES, or LINK authorization method.
3. Create one pending application for one of the two partners.
4. Check that the `Apply` button for the partner with the pending application is disabled.
5. Check that the `Apply` button for the partner without the pending application is enabled.

## Screenshots of your changes (if appropriate):
[//]: # (It can also be a GIF to prove that your changes are working)
**Partner with Authorization Method EMAIL that has no open applications**
Screen Shot 2021-03-12 at 21 36 39

**Partner with Authorization Method EMAIL that has an open application**
Screen Shot 2021-03-12 at 21 36 55

**My Library view with the Extend and View Application buttons showing**
Screen Shot 2021-03-15 at 19 55 07

## Types of changes
What types of changes does your code introduce? Add an `x` in all the boxes that apply:
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

PR #650, opened 2021-03-16
HTML string has been made translatable

sahilgrewal8072:
Following file TWLight/users/templates/users/resource_tile.html has been modified and added {% trans 'text goes here' %}
on the requested line 111 also comment has been added above the given line.

Bug: T277155

### DIFF
--- a/TWLight/users/templates/users/resource_tile.html
+++ b/TWLight/users/templates/users/resource_tile.html
@@ -108,7 +108,8 @@
       </div>
       <hr>
     {% endif %}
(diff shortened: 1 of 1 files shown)


suecarmol:
Hi @sahilgrewal8072, thank you for your contribution. Since this file was updated recently In #648, could you please rebase your branch against the master branch to update the file?

suecarmol (approved):
Looks good! I thought that with merging #648, there would be a conflict, but this doesn't seem to be the case. Thanks for contributing to The Wikipedia Library

PR #654, opened 2021-03-24
Add management command to add translatable partner descriptions

suecarmol:

## Description
Created a management command to transfer all the partner descriptions to JSON files so they can be translatable on TranslateWiki.

This is PR 1/3 to address this.

## Rationale
Partner descriptions couldn't be translated because they were created in the Django admin interface.

## Phabricator Ticket
[T273877](https://phabricator.wikimedia.org/T273877)
[T278334](https://phabricator.wikimedia.org/T278334)

## How Has This Been Tested?
Manually tested the command and checked that it was creating well-formatted JSON files.

## Types of changes


Samwalton9:
We occasionally rename partners, will that still work in the way this is set up?

suecarmol:
> We occasionally rename partners, will that still work in the way this is set up?

That's a great point. I have changed the partner name to the partner key (which is immutable) so that we don't face those issues.

jsnshrmn (commented):
This looks really good to me. I've just reached out to the translatewiki folks to get their take on the way we're organizing this to make sure they are happy with it.
https://translatewiki.net/w/i.php?title=Support&lqt_method=talkpage_new_thread
I really should have done this last week, so sorry for stalling you out!

jsnshrmn:
Okay, from the discussion with translatewiki thus far:

- We should lint the json to make sure we're strictly valid. I'm happy to add that into our CICD. If it's okay with you, I can just add that into this PR so we can see it in play here
- We should encode the messages with UTF-8 instead of using ASCII escape sequences
- We might need to drop all of the empty messages. That bit is under discussion.

jsnshrmn:
Another option would be to use translatewiki as our interface for adding those

jsnshrmn:
> We could also use JSON schema to validate our JSONs.
Even better, so we can just add a unit test that runs that against the translations.

suecarmol:
> Another option would be to use translatewiki as our interface for adding those

Ok, so then I'll delete the empty JSON files and leave the others :)

suecarmol:
> > We could also use JSON schema to validate our JSONs.
> 
> Even better, so we can just add a unit test that runs that against the translations.

On it, then!

jsnshrmn:
> 
> 
> > Another option would be to use translatewiki as our interface for adding those
> 
> Ok, so then I'll delete the empty JSON files and leave the others :)

👍 
When we talked about updating the messages in code, we kind of put a pin in talking about what the whole process would look like. If this follows the same flow as our ugettext stuff, `en` and `qqq` will be outputs from us and inputs for wikitranslate; all the other translations will be outputs from wikitranslate and inputs for us. There shouldn't be a problem with them ingensting the existing translations that first time around though.

suecarmol:
> > > Another option would be to use translatewiki as our interface for adding those
> > 
> > 
> > Ok, so then I'll delete the empty JSON files and leave the others :)
> 
> 👍
> When we talked about updating the messages in code, we kind of put a pin in talking about what the whole process would look like. If this follows the same flow as our ugettext stuff, `en` and `qqq` will be outputs from us and inputs for wikitranslate; all the other translations will be outputs from wikitranslate and inputs for us. There shouldn't be a problem with them ingesting the existing translations that first time around though.

Got it! I have deleted all of the files except for `en` and `qqq`. Although we have translations from other languages, we can add them manually via WikiTranslate

jsnshrmn (approved):
LGTM! I'll let the translatewiki folks know we're ready for them to include this for translation.

verdy-p:
: "If this follows the same flow as our ugettext stuff, /en and /qqq will be outputs from us and inputs for wikitranslate"

Ok for "/en", but the "/qqq" is an editable area on Translatewiki.net, where feedback of additional guiding is submitted. If you overwrite them blindly, this would erase these.

I would suggest separating the imports currently made directly on"/qqq" from the editable "/qqq" page, so that these imports would be done e.g. on "/qqi" (like "info") which would be uneditable, but that an still be rendered on the "/qqq" page by local transclusion (it may be removed if inappropriate, such as incorrect links to sources: a frequent error).

The "/qqq" contains other useful tracking templates for handling coherent terminologies (e.g {{Identical|...}}, and other "see also" and relevant links to talks or past decisions made elsewhere than TWN itself). It should remain usable, viewable and editable directly in TWN.

If needed, Translatewiki.net would just add an "unrollable section" or just a link to the imported "/qqi" (if one was imported), but otherwise would protect "/qqi" exactly like the source language subpage (most often "/en"). The "/qqi" subpage could also contain other metadata (such as a common link to the project support page or repository, or metadata about the syntax or language used to parsee the original string). It may contain some docs.

When importing a "/qqi" page, the import tool would just check if the "/qqq" page does not exist or if if exists but does not contain transcludes or links the "/qqi" page, and so it may create that "/qqq" page or would update it if there's no such transclusion (that transclusion can still be hidden with surrounding markup in the qqq page, such as HTML comments when it is not relevant: that't why I propose instead that TWN.net allows showing a link, or tabs to switch between the "/qqq" or "/qqi" view, or a button to hide/unhide the "/qqi" page, which may be more technical and managed by import bots).

jsnshrmn:
Thanks for the info on `qqq`. It shouldn't be an issue for these json files, since we don't have any tooling that updates it automatically. As long as the translatewiki `qqq` documentation is getting committed, we shouldn't need to touch it. We'll just update `en` when we get new partners or their description needs changed.

verdy-p:
I still sponsor the idea of a JSON-formatted "/qqi" page for use be import bots: it should be a keyed dictionnary (i.e. an associative array with string keys), and TWN should reserve some keys starting by "_" for future standardisation; their value could be for example the flags using in special comments of .po files for gettext (in lines starting by "#," just before the lines with "msgid" and "msgstr" for the source identifier or message in English and the translated text).

Some other keys could be reserved to point another page containing global rules imported as defaults for the whole set of translatable resources, such as plural rules, or default programming language or syntax)

Immediately import bots would be free to update the "/qqi" pages themselves, and these won't be editable on TWN's UI but will be usable by it to better hint its validator and guide users, that will continue to edit translations in other "/<lang>" pages and their doc in "/qqq". I just suggest that these qqi pages should be easily parsable, and so should be a JSON object, an associative array indexed by string keys).

Then TWN can develop itself independantly, and will also be able to support more projects than those it currently supports (still with lot of false alarms due to false guesses), and so it could accept simple imports from other translation tools and eaiser export formats to development repository.

jsnshrmn:
Are there some projects using this arrangement that I can go look at as an example?

verdy-p:
No, that's just something that I propose (and that should be discussed with TWN's admins, to organize the storage arrangement, authorizations for bots, content-models, ... and later develop TWN to handle such improvements in the validation and maintenance). It could still be tested independantly of TWN's development of its UI (to take into account this data for its own local validator and enhance the translate interface).

The translate interface is co-developed also within Wikimedia for translated pages on its wikis, as it uses the same Mediawiki UI extension). One problem in TWN is that it still does not support features supported in Wikiemdia, notably Scribunto with Lua modules; and the version of Mediawiki deployed may lag behind.

PR #656, opened 2021-03-27
T278420

sahilgrewal8072:
Changes have been made to hide red block message when the user is whitelisted

File TWLight/templates/home.html have been modified successfully for the above update

Bug:T278420

### DIFF
--- a/TWLight/templates/home.html
+++ b/TWLight/templates/home.html
@@ -68,7 +68,7 @@
             </li>
           {% endfor %}
         </ul>
-        {% if user.is_authenticated and not user.editor.wp_not_blocked %}
+        {% if user.is_authenticated and not user.editor.ignore_wp_blocks and not user.editor.wp_not_blocked %}
           <p class="block-notice">
             {% url 'contact' as contact_url %}
             {% comment %}Translators: This message is shown on the homepage to users whose Wikimedia account is blocked on one or more projects. They may still be granted access to the Library Bundle, but need to contact Wikipedia Library staff to verify. {% endcomment %}


TWLight/templates/home.html
```diff
@@ -68,7 +68,7 @@ <h2 style="padding-bottom:20px;">{% trans "Library Bundle" %}
             </li>
           {% endfor %}
         </ul>
-        {% if user.is_authenticated and not user.editor.wp_not_blocked %}
+        {% if user.is_authenticated and not user.editor.ignore_wp_blocks and not user.editor.wp_not_blocked %}
```
suecarmol, line 71: As far as I understand, the `ignore_wp_blocks` flag should be on to ignore the blocks, so this condition has to be: 

```suggestion
        {% if user.is_authenticated and user.editor.ignore_wp_blocks and not user.editor.wp_not_blocked %}
```

@Samwalton9 is this correct?

suecarmol (changes requested):
Thank you for your contribution and apologies for the late review. I think this condition needs to be changed in order for this to get merged.

TWLight/templates/home.html
Samwalton9: This line is a little confusing because the block field is `not_blocked` so we're looking at `not not_blocked` (blocked) to see if we should show this message. It decides whether we show the 'you are blocked message' and we only want to do that if we _haven't_ ignored their WP blocks.

suecarmol (approved):
Based on @Samwalton9's comment, this is correct. Moving to merge

PR #677, opened 2021-04-24
Create tag names files.

suecarmol:
## Description
Create `en` and `qqq` files for tags so TranslateWiki can pick them up.

## Rationale
This is the first step that will make tag names translatable.

## Phabricator Ticket
[T281013](https://phabricator.wikimedia.org/T281013)

## How Has This Been Tested?

## Types of changes


locale/en/tag_names.json
```diff
@@ -0,0 +1,39 @@
...
+			"Susana Cárdenas Molinar",
+			"Jason Sherman",
+			"Sam Walton"
+		],
+		"locale": "en",
+		"message-documentation": "qqq"
+	},
+    "language_translation": "language",
+    "multidisciplinary_translation": "multidisciplinary",
+    "physical-science_translation": "physical-science",
```
Samwalton9, line 13: I think these tags had no spaces/capital letters in the past for technical reasons. I think they could all be properly capitalised and have hyphens removed in favour of spaces (so in this case **Physical science**).

locale/en/tag_names.json
```diff
@@ -0,0 +1,39 @@
...
+    "music_translation": "music",
+    "art_translation": "art",
+    "social_translation": "social",
+    "sciences_translation": "sciences",
```
jsnshrmn, line 17: I think that these two tags exist by mistake, and that the one resource that has them should probably be using the `social-sciences` tag.

jsnshrmn (commented):
seems like our message keys should be different? instead of `art_translation` maybe we should have `art_tag` or `tag_art`?

locale/en/tag_names.json
suecarmol: Done!

locale/en/tag_names.json
suecarmol: Those message keys make more sense. Have removed both tags.

jsnshrmn (approved):
👍

PR #680, opened 2021-04-26
Reorder partner descriptions to make it easier to translate

amire80:

## Description
1. Put long descriptions after short descriptions.
2. Put all the items in the same order in which they appear on the website.

## Rationale
1. The short description is an introduction, and the long one, when it exists, is easier to understand when it appears after the intro.
2. As I was translating the description, it was easier to also look at the website, to understand how they will appear to the users. Seeing them in the same order as on the website would be more convenient.

## Types of changes
- [x] Minor change (fix a typo, add a translation tag, add section to README, etc.)


jsnshrmn (approved):
A reasonable change. Thanks!

PR #684, opened 2021-04-28
Remove `move_partner_description_to_files` command

suecarmol:
## Description
Removed the `move_partner_description_to_files` management command.

## Rationale
This command is no longer needed since the partner descriptions JSON files will be updated directly from the source code (in the case of the `en` and `qqq` files) or from TranslateWiki (the rest of the localization files).

We also need to remove this command before we remove the description columns from the `Partner` model, otherwise running this command will result in errors.

Also, less code to worry about!

## Phabricator Ticket
[T281388](https://phabricator.wikimedia.org/T281388)

## How Has This Been Tested?
This had no tests and removing this does not affect any other parts in the codebase.

## Types of changes
- [] Minor change (fix a typo, add a translation tag, add section to README, etc.)


jsnshrmn (approved):
👍

PR #703, opened 2021-05-10
Improve pip build speed

jsnshrmn:

## Description
Removes some unnecessary dependencies and enables pip caching. Stops force installing things in the venv that are already provided upon venv creation. Sets a fixed version on a dependency that had no version specified.

## Rationale
When any one of our pip dependencies change, we have to reinstall the whole set from scratch. Some of them require compilation, which is quite time consuming. This is exacerbated in our cicd environment where there is no local build cache

## Phabricator Ticket
https://phabricator.wikimedia.org/T282456

## How Has This Been Tested?
I've run this build locally many times to check functionality and speed.
I also updated the production twlight_build image in the registry with the pip cache and verified that I could make use of it with an empty local build cache

Before:
```
Successfully built ea1c249329c6ec589896afa3d43481acb00b9cb3bd4408662a05cdb18c840a83

real	7m25.785s
user	0m28.105s
sys	0m16.432s
```
After:
```
Successfully built bc6a79e2ff6f86204b9f2c67f238df2c253e2da161b84b29529f4c875f80e2a2

real	3m22.031s
user	0m23.408s
sys	0m11.188s
```

## Types of changes


suecarmol (approved):
Thanks for this! Anything that helps speed up our build times is a welcome change

PR #713, opened 2021-05-14
exclude bots from Matomo analytics

jsnshrmn:
## Description
Adds more user agents to the bot list and filters all bot traffic out from matomo analyitcs

## Rationale
We're trying to improve the signal:noise ratio in matomo

## Phabricator Ticket
https://phabricator.wikimedia.org/T282714

## How Has This Been Tested?
I verified that this works locally by connecting my dev environment to matomo and setting the user-agent string in curl.
I'll need to update the visitor segments in matomo, because the `Bot` and `Not-Bot` variables are used in those, but will no longer be populated.

N/A

## Types of changes


suecarmol (approved):
Thanks for this!

PR #720, opened 2021-05-27
Redevelop tags

suecarmol:
## Description
Redevelop tagging system for subject classification, update tests to pass for the new tagging system, and update partners' tags to reflect the changes.

## Rationale
Address inconsistencies in previous tagging system (eg format- vs subject-based tags)

## Phabricator Ticket
[T283857](https://phabricator.wikimedia.org/T283857)

## How Has This Been Tested?
Ran the management command manually without any errors (the `new_tags` JSON was validated successfully).

## Types of changes


jsnshrmn (commented):
This works alright and the code is clean, but I don't love that it would leave the browse partners page in an error state until the management command gets run. I'd recommend wrapping the partners in a guard clause, like
```
        partner_24 = Partner.objects.get(pk=24, company_name="AAAS")
        if partner_24:
                partner_24.new_tags = {
                    "tags": [
                        "life-sciences_tag",
                        "physical-sciences_tag",
                        "health-sciences_tag",
                        "earth-sciences_tag",
                    ]
                }
                partner_24.save()
```
and then calling the command in a migration. That way, it runs on deploy, but is effectively a noop if those partners aren't there.

Having said all that, the code does work as is, and we don't currently have a huge userbase:
if you want to shell in and watch the deploy, I'm sure you could fire the command within a few seconds and have a pretty small window for folks to encounter issues. If that's what you'd like to do, feel free to merge!

jsnshrmn:
Another option would be to drop the management command code directly into a migration

suecarmol:
I have added the migration file to run the command and added a guard clause.

jsnshrmn:
I think the build failure is just down needing to rebase to grab a missing migration and update your migration dependency.

suecarmol:
> I think the build failure is just down needing to rebase to grab a missing migration and update your migration dependency.

Yup, fixed that, CI is running again

jsnshrmn:
Ah, yeah, swapping those `get`s for `filter().first()` will get you a `None` when there is no such object.

suecarmol:
> Ah, yeah, swapping those `get`s for `filter().first()` will get you a `None` when there is no such object.

On it!

jsnshrmn (approved):
We are go for launch!
🚀🚀🚀

PR #750, opened 2021-07-12
Die Zeit -> ZEIT ONLINE

Samwalton9:
Updating Die Zeit to the correct name for the online database


jsnshrmn (commented):
It seems like the qqq documentation for the partner should change too?
https://github.com/WikipediaLibrary/TWLight/blob/2fbf14ca7b98268b24e4794cc9d4496c57713fd6/locale/qqq/partner_descriptions.json#L28

Samwalton9:
Oh yep thanks, updated.

jsnshrmn (approved):
😎

PR #756, opened 2021-07-16
Fix navbar options when unauthenticated

suecarmol:
To show the Log Out and Profile/Settings option in the header

## Description
Add a clause to check if a user is authenticated in order to show some navbar options like `Log out` and `Profile/Settings`

## Rationale
It doesn't make sense to show these options when a user is logged out of the application.

## Phabricator Ticket
[T286734](https://phabricator.wikimedia.org/T286734)

## How Has This Been Tested?
1. Log out of TWL
2. Go to a page where a user can go unauthenticated, like the `Contact Us` page.
3. Check that the `Log out` option does not show

**Before the fix**
Screen Shot 2021-07-15 at 19 23 43

**After the fix**
Screen Shot 2021-07-15 at 19 23 56

## Types of changes

### DIFF
--- a/TWLight/templates/header_partial_b3.html
+++ b/TWLight/templates/header_partial_b3.html
@@ -50,20 +50,19 @@
           </a>
         </li>
         <li role="separator" class="divider"></li>
-        {% if user %}
-        <li role="presentation" class="mobile-menu nav-item">
-          <a class="btn user-button" href="{% url 'users:home' %}">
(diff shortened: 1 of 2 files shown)


jsnshrmn (approved):
Oopsie! This is a straightforward fix, and it looks good. I think we should probably add a test for authenticated vs anonymous nav content, but I don't think that should block this bugfix.

PR #762, opened 2021-07-20
Implement new collection tiles

suecarmol:
[//]: # (Thank you for uploading a PR to the Wikipedia Library!)

## Description
Redesign collection tiles in the My Library UI.

## Rationale
[//]: # (Why is this change required? What problem does it solve?)
This task is part of our library redesign that aims to make TWL more user-friendly.

## Phabricator Ticket
[//]: # (Link to the Phabricator ticket)
[T285552](https://phabricator.wikimedia.org/T285552)

## How Has This Been Tested?
[//]: # (- Did you add tests to your changes? Did you modify tests to accommodate your changes?)
[//]: # (- Can this change be tested manually? How?)

## Screenshots of your changes (if appropriate):
[//]: # (It can also be a GIF to prove that your changes are working)

**Future `My Collections` section**
Screen Shot 2021-07-22 at 18 13 44

**Future `Available Collections` section**
Screen Shot 2021-07-22 at 18 13 57

## Types of changes
What types of changes does your code introduce? Add an `x` in all the boxes that apply:
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Minor change (fix a typo, add a translation tag, add section to README, etc.)

PR #767, opened 2021-07-27
Implement new My Library UI

suecarmol:

## Description
Implement the missing pieces of My Library UI after completing #762. 

## Rationale
This is part of TWL redesign efforts.

## Phabricator Ticket
[T285556](https://phabricator.wikimedia.org/T285556)

## How Has This Been Tested?
Tested manually and created tests for the filters

Screen Shot 2021-07-27 at 11 55 20

## Types of changes

### DIFF
--- a/TWLight/users/templates/users/filter_section.html
+++ b/TWLight/users/templates/users/filter_section.html
@@ -10,7 +10,7 @@
     <div class="row filter-title-container">
       <p class="filter-title">
         {% comment %}Translators: On the MyLibrary page (https://wikipedialibrary.wmflabs.org/users/my_library), this label indicates the start of the filter section. {% endcomment %}
-        {% trans "Filters: " %}
+        {% trans "Filters" %}:
       </p>
       <a href="{% url 'users:redesigned_my_library' %}?tags=&languages=" class="reset-filters-link">
         {% comment %}Translators: On the MyLibrary page (https://wikipedialibrary.wmflabs.org/users/my_library), this link will reset the filters for the collections. {% endcomment %}
--- a/TWLight/users/templates/users/filter_section.html
+++ b/TWLight/users/templates/users/filter_section.html
@@ -1,0 +1,67 @@
+{% load i18n %}
+<nav class="navbar navbar-expand-lg navbar-light">
+  <button class="navbar-toggler" type="button" data-toggle="collapse"
+      data-target="#filterNavbar" aria-controls="filterNavbar" aria-expanded="false"
+      aria-label="Toggle navigation">
+    <span class="navbar-toggler-icon"></span>
+  </button>
+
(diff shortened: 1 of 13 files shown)


jsnshrmn:
Okay, I've manually tested emulating several devices (phone, tablet, desktop) and several languages (including hebrew, which is rtl) and it's functionally good IMO, other than the aforementioned issue with filtering languages. Moving on to a readthrough.

suecarmol:
> The first thing I've noticed is that the language select filter has some odd behavior. On a fresh load/ hard reload (shift + F5) I find it almost impossible to select a language with a mouse click. Clicking left mouse also scrolls the list down and selects whichever language is beneath the cursor when the button is released. I'm able to (unreliably) select a language with the mouse once I've already fiddled with the control and there is some state.

I do not see this behavior in Chrome or Firefox for Mac. Do you have more details about how to reproduce this error?

jsnshrmn:
It's very consistent for me on firefox 90 for ubuntu on this view, but I'm not seeing it on firefox 90 for windows or android on staging.

jsnshrmn:
I will do some more testing to try to reproduce this in a clean  Linux vm to verify that it's not something weird about my system.

jsnshrmn:
We verified that this was a firefox 90.0 bug that's resolved in the current release, 90.02.

TWLight/users/templates/users/filter_section.html
```diff
@@ -0,0 +1,51 @@
...
+      data-target="#filterNavbar" aria-controls="filterNavbar" aria-expanded="false"
+      aria-label="Toggle navigation">
+    <span class="navbar-toggler-icon"></span>
+  </button>
+
+  <div id="filterNavbar" class="collapse navbar-collapse filter-container">
+    <div class="row filter-title-container">
+      <p class="filter-title">
+        {% comment %}Translators: On the MyLibrary page (https://wikipedialibrary.wmflabs.org/users/my_library), this label indicates the start of the filter section. {% endcomment %}
+        {% trans "Filters: " %}
```
jsnshrmn, line 13: I think we should not be including the `: ` in our strings for translation. We had a couple of these sneak in over the last pr or two in `available_collection_tile.html` and `user_collection_tile.html` and we should bring these all into line with what we're doing elsewhere, which is like `{% trans "Filters" %}: `

jsnshrmn (changes requested):
This looks great! I have one inline nitpick on localization strings, but other than that this is good to go IMO.

TWLight/users/templates/users/filter_section.html
suecarmol: Done!

suecarmol:
I also fixed the `:` from the available collections and user collections templates

jsnshrmn (approved):

PR #771, opened 2021-08-02
Fix More Info bug that reverted text to English

suecarmol:

## Description
Fixed a bug where the `More Info` button would revert back to English instead of a user's chosen language when the button was clicked again.

## Rationale
Fixes a bug that was created a less enjoyable experience for non-English users.

## Phabricator Ticket
[T287577](https://phabricator.wikimedia.org/T287577)

## How Has This Been Tested?
Tested manually in the interface

**Before**
Before_fix

**After**
After_fix

## Types of changes

### DIFF
--- a/TWLight/templates/homepage.html
+++ b/TWLight/templates/homepage.html
@@ -334,6 +334,8 @@
     // Getting the "More info" translation from the first button element
     var partnerButtons = document.getElementsByClassName("partner-glider-button");
     var partnerText = partnerButtons[0].innerText;
+    // Getting the Less info text from the view
+    var lessInfo = "{{ less_info }}";
 
     function toggleDescription(partnerId) {
       // Before toggling the description, search if there are any descriptions
@@ -373,7 +375,7 @@
       // If the invisible class exists in the partner image, then the button
       // needs to show "Less info"
       if (partnerImage.classList.contains("invisible")) {
-        partnerButton.innerHTML = "Less info";
+        partnerButton.innerHTML = lessInfo;
       } else {
         partnerButton.innerHTML = partnerText;
       }
--- a/TWLight/templates/homepage.html
+++ b/TWLight/templates/homepage.html
@@ -331,6 +331,10 @@
     $(function () {
       $('[data-toggle="tooltip"]').tooltip()
     });
+    // Getting the "More info" translation from the first button element
+    var partnerButtons = document.getElementsByClassName("partner-glider-button");
+    var partnerText = partnerButtons[0].innerText;
+
     function toggleDescription(partnerId) {
       // Before toggling the description, search if there are any descriptions
       // already toggled and flip them to the image
@@ -347,7 +351,6 @@
           partnerContainer.classList.toggle("partner-description-toggled");
         }
       }
-
       toggleElement("partner-image-" + partnerId);
       toggleElement("partner-description-" + partnerId);
       togglePartnerButton(partnerId);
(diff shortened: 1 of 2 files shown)


Samwalton9:
From looking at the gifs am I right in thinking that the 'Less info' button still displays in English?

suecarmol:
> From looking at the gifs am I right in thinking that the 'Less info' button still displays in English?

Yes, `Less Info` does not have a translation tag because it is originated from javascript (see [this line])(https://github.com/WikipediaLibrary/TWLight/blob/fa8be46a8d4cf38bfe3916b1fdd4884e7df9f504/TWLight/templates/homepage.html#L376)

suecarmol:
We could replace `More info` and `Less info` with `Flip`.

Samwalton9:
I think Flip is a lot less clear. Is there no way to mark text in JS as translatable?

suecarmol:
> I think Flip is a lot less clear. Is there no way to mark text in JS as translatable?

Looks like we can! I passed the `Less info` string from the view so it can be translatable.

TWLight/templates/homepage.html
```diff
@@ -370,9 +375,9 @@
       // If the invisible class exists in the partner image, then the button
       // needs to show "Less info"
       if (partnerImage.classList.contains("invisible")) {
-        partnerButton.innerHTML = "Less info";
+        partnerButton.innerHTML = lessInfo;
```
jsnshrmn, line 378:
```suggestion
        {% comment %}Translators: This text will show when a user clicks on the More Info button and the partner description shows.{% endcomment %}
        partnerButton.innerHTML = "{% trans 'Less info' %}";
```

TWLight/templates/homepage.html
```diff
@@ -370,9 +375,9 @@
...
       } else {
-        partnerButton.innerHTML = "More info";
+        partnerButton.innerHTML = partnerText;
```
jsnshrmn, line 380:
```suggestion
        {% comment %}Translators: A button that reveals more information about a collection.{% endcomment %}
        partnerButton.innerHTML = "{% trans 'More info' %}";
```

TWLight/templates/homepage.html
```diff
@@ -331,6 +331,12 @@
     $(function () {
       $('[data-toggle="tooltip"]').tooltip()
     });
+    // Getting the "More info" translation from the first button element
+    var partnerButtons = document.getElementsByClassName("partner-glider-button");
+    var partnerText = partnerButtons[0].innerText;
+    // Getting the Less info text from the view
+    var lessInfo = "{{ less_info }}";
+
```
jsnshrmn, line 339:
```suggestion

```

TWLight/views.py
```diff
@@ -181,6 +181,9 @@ def get_context_data(self, **kwargs):
             )
         context["partners"] = partners_obj
 
+        # Translators: This text will show when a user clicks on the More Info button and the partner description shows.
+        context["less_info"] = _("Less info")
+
```
jsnshrmn, line 186:
```suggestion

```

jsnshrmn (changes requested):
We should be able to handle all of this in the template, since that's where the strings live.

suecarmol:
Thanks for suggesting these changes! It makes the code more legible

jsnshrmn (approved):

jsnshrmn:
thanks!

PR #773, opened 2021-08-03
Renamed instances of Stream with Collection

aacaldwell:
Bug: T275040

## Description

## Rationale
[//]: # Notes from the Phabricator ticket
Some publishers in the library have distinct collections of content that users can apply for. In the Library Card platform we sometimes refer to these as 'streams' and sometimes as 'collections'. This is confusing both for users and for folks translating the tool.

We should update strings (but not variable names for now, for simplicity) anywhere we've used the word 'stream' to instead say 'collection'. These strings might be in python code or in templates.

## Phabricator Ticket
[//]: # https://phabricator.wikimedia.org/T275040

## How Has This Been Tested?

I did not add any changes to the test. I did run docker-compose exec twilight /app/bin/virtualenv_test.sh to make sure everything passed, but I may have missed something because of the test. However, it should not be affected since only the strings were changed, not the variables.

## Types of changes

Not sure how to classify this.

### DIFF
--- a/TWLight/applications/helpers.py
+++ b/TWLight/applications/helpers.py
@@ -28,7 +28,7 @@ Optional/universal:
 Optional/unique:
     Questions/comments/concerns (free-text, all partnerships)
     Title requested (McFarland, Pelican)
-    Stream requested (OUP, T&F, Elsevier)
+    Collection requested (OUP, T&F, Elsevier)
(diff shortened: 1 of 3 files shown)


Samwalton9:
Per my note at https://phabricator.wikimedia.org/T275040.

PR #780, opened 2021-08-05
Launch MVP

suecarmol:

## Description
- Removed old My Library view and template
- Replaced old My Library view with the redesigned one
- Changed behavior on user login: when a bundle-eligible user logs in, they are redirected to My Library instead of the home screen
- Changed `Go back to partners` button 

## Rationale
This is all part of the launch of the redesigned My Library MVP.

## Phabricator Ticket
[T285558](https://phabricator.wikimedia.org/T285558)

## How Has This Been Tested?
Tested manually

**Login User is not Bundle-Eligible**
LoginHome

**Login User is Bundle-Eligible**
LoginMyLibrary

## Types of changes


TWLight/users/tests.py
```diff
@@ -357,162 +357,6 @@ def test_sent_application(self):
         app.refresh_from_db()
         self.assertNotIn("Withdraw", response.render().content.decode("utf-8"))
 
-    def test_my_library_authorizations_and_collections(self):
```
jsnshrmn, line 360: It looks like we may be losing some authorization test coverage here? Maybe instead of dropping this entirely, you could revert the change I made here:
https://github.com/WikipediaLibrary/TWLight/commit/160fcee59bb97eecc696e7d66082c942019ed23a#diff-bdbc11d651f956ef2d399ac5ec4d943b6ede2018b9df52b37b465a60e8752e8e

So that this test is really only concerned with the authorizations themselves.

This was just based on a read through, so I could be mistaken. Happy to circle back around for another look on Tuesday.

suecarmol: I think the change you added to the test is the only thing we should keep, since we don't make distinctions between manual authorizations and bundle authorizations anymore.

jsnshrmn: This was a red herring in my part. As you say, we don't care about some of the differences in the auths anymore.

jsnshrmn (approved):
Looks good to me! Thanks!

PR #783, opened 2021-08-08
Rename duplicate test function names

akayjoshi:
## Description
Renamed duplicate test functions with different code
and Removed duplicate test with same code

## Phabricator Ticket
https://phabricator.wikimedia.org/T282460

## Types of changes

### DIFF
--- a/TWLight/applications/tests.py
+++ b/TWLight/applications/tests.py
@@ -1840,7 +1840,7 @@ class ListApplicationsTest(BaseApplicationViewTest):
                 escape(obj.__str__()), response.render().content.decode("utf-8")
             )
 
-    def test_list_object_visibility(self):
+    def test_list_deleted_object_visibility(self):
         url = reverse("applications:list")
         queryset = Application.objects.filter(
             status__in=[Application.PENDING, Application.QUESTION]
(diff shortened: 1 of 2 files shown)


TWLight/applications/tests.py
```diff
@@ -1840,23 +1840,14 @@ def _base_test_deleted_object_visibility(self, url, view, queryset):
...
             )
 
-    def test_list_object_visibility(self):
-        url = reverse("applications:list")
-        queryset = Application.objects.filter(
-            status__in=[Application.PENDING, Application.QUESTION]
-        )
-        self._base_test_deleted_object_visibility(
-            url, views.ListApplicationsView, queryset
-        )
```
jsnshrmn, line 1850: I'm not actually seeing where else this queryset is being run against `_base_test_deleted_object_visibility`; this looks to me like we'd be losing test coverage.

jsnshrmn (changes requested):
Thanks for your work on this! The method renames look good, but I do have an inline comment on the method you deleted.

TWLight/applications/tests.py
akayjoshi: yes you were right I mistakenly deleted that thinking its duplicate, I have rewrote it now

jsnshrmn: Thanks! These method names are a big improvement in descriptiveness in addition to removing the naming collisions.

jsnshrmn (approved):

PR #800, opened 2021-08-28
Change internal server error to admin validation error

akayjoshi:
## Description
When an admin enters an invalid tag for a resource, a ValidationError is thrown, presenting the user with a server error message
Instead, it should simply present a red message next to the relevant field, denoting why the object cannot be saved.

## Phabricator Ticket
https://phabricator.wikimedia.org/T287003

## Types of changes


akayjoshi:
why Travis didn't check this :thinking:

TWLight/resources/migrations/0080_auto_20210828_0548.py
```diff
@@ -0,0 +1,23 @@
...
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="partner",
+            name="new_tags",
+            field=models.JSONField(
+                blank=True,
+                default=None,
+                help_text="Tag must be a valid JSON schema. Tag should be in the form of TagName_tag.",
```
jsnshrmn, line 19:
```suggestion
                help_text="Tag must be a valid JSON schema. Tag should be in the form of tag-name_tag.",
```

TWLight/resources/models.py
```diff
@@ -279,7 +279,13 @@ class Meta:
     )
 
     # New tag model that uses JSONField instead of Taggit to make tags translatable
-    new_tags = models.JSONField(null=True, default=None, blank=True)
+
+    new_tags = models.JSONField(
+        null=True,
+        default=None,
+        blank=True,
+        help_text="Tag must be a valid JSON schema. Tag should be in the form of TagName_tag.",
```
jsnshrmn, line 287:
```suggestion
        help_text="Tag must be a valid JSON schema. Tag should be in the form of tag-name_tag.",
```

TWLight/resources/models.py
```diff
@@ -423,6 +424,9 @@ def save(self, *args, **kwargs):
                     "Error trying to insert a tag: the JSON is invalid"
```
jsnshrmn, line 424: For bonus points (not a blocker for merging) Since we're making this more usable, let's give the person a link to the list of available tags. Note that you'll also need to import mark_safe, eg
`from django.utils.safestring import mark_safe`
and update the tests.

```suggestion
                    mark_safe("Error trying to insert a tag: please choose a tag from <a rel='noopener' target='_blank' href='https://github.com/WikipediaLibrary/TWLight    /blob/production/locale/en/tag_names.json'>tag_names.json</a>.")
```

jsnshrmn (changes requested):
Thanks for your work on this! I'll investigate why travis didn't run, but in the mean time, I locally pulled your code and rebased it to do automated and manual testing. It works great, and the change is straightforward. I have inline suggestions that you can directly commit for the help text, which is a little off in the suggested tag form. I also have a bonus suggestion that you don't have to implement.

TWLight/resources/models.py
akayjoshi: Hello @jsnshrmn, thanks for the suggestion I will work on it. I was also thinking about this when testing code manually because I don't know what is valid tags that I can give for testing functionality.

akayjoshi: Hello @jsnshrmn, I have updated required changes. Thank you.

jsnshrmn: :+1:

jsnshrmn (approved):

jsnshrmn (commented):
I noticed that this PR included an auto migration, which we exclude to allow travis to create them. This helps avoid migration conflicts. I went ahead and deleted the file from the PR and am just waiting for tests to run before merging.
Thanks for your work!

akayjoshi:
Sorry, I didn't know about this.

PR #803, opened 2021-09-01
Stream migration delete streams and create new partners

suecarmol:

## Description
* Created new partners based on streams
* Created a JSON file that temporarily holds stream descriptions
* Reassigned applications made on streams to newly created partners
* Reassigned authorizations made on streams to newly created partners
* Deleted streams

## Rationale
We are in the process of migrating the last of our streams (also sometimes referred to as collections) into individual resources (now referred to as collections). This won't make the process of browsing the library much more confusing for users, but will allow us to finally deprecate streams entirely.

This is also a blocker to upgrade to Django 3.2.

## Phabricator Ticket
[T289283](https://phabricator.wikimedia.org/T289283)

## How Has This Been Tested?
Ran the command manually.

## Types of changes


jsnshrmn (changes requested):
I pulled this down and ran it locally against a recent db dump. It deleted the streams and partners, but did not create the new partners, so there's something going on there. No exceptions were raised nor errors logged.

suecarmol:
> I pulled this down and ran it locally against a recent db dump. It deleted the streams and partners, but did not create the new partners, so there's something going on there. No exceptions were raised nor errors logged.

Thanks for catching that! Another reason why tests are needed (even though we will only be running this command once).

jsnshrmn (changes requested):
Okay, now we've got the new partners! A few things that I noticed:
- We should not be setting the `specific_stream` toggle on the new partners since there are no streams now. This breaks the application form for those partners.
- Ideally we'd preserve all of the existing dates for applications and authorizations. Right now the state ends up looking like we approved  a bunch of apps the moment we run the command, and all the expiration dates are stretched out accordingly.

suecarmol:
> Ideally we'd preserve all of the existing dates for applications and authorizations. Right now the state ends up looking like we approved a bunch of apps the moment we run the command, and all the expiration dates are stretched out accordingly.

What dates are being changed in applications? I think in authorizations, `date_authorized` is being changed because it has an `auto_now_add=True` flag. The expiration dates are being changed only if the authorization has already expired. I'll take a look at that.

jsnshrmn:
> What dates are being changed in applications? I think in authorizations, `date_authorized` is being changed because it has an `auto_now_add=True` flag. The expiration dates are being changed only if the authorization has already expired. I'll take a look at that.

It looks like a new revision is being created; which you can see in the finalized tab of the review page.
From a run earlier this morning:
image

jsnshrmn:
Probably the most straightforward way to change *just* the related partners and streams for apps and auths  would be to fallback to raw sql for the updates.

suecarmol:
> Probably the most straightforward way to change _just_ the related partners and streams for apps and auths would be to fallback to raw sql for the updates.

Got it!

jsnshrmn (commented):
The code looks good. When I run it, the authorizations look good. I noticed that the total number of applications drops from 11837 to 11748, a difference of 89. I've been poking around a bit, but haven't yet determined why this would be the case.

jsnshrmn:
I went through and counted up various groupings of applications before and after running the command, hoping to find an obvious commonality.  All the groups lost some, but future science lost the most

````
before partner:
future science: 62
RILM: 48
Springer Nature: 169

before status:
pending: 401
under discussion: 11
approved: 50
not approved: 2148
sent: 9012
invalid: 215

after partner:
future science: 8 + 4 = 12
RILM: 19 + 14 = 33
Springer Nature: 81 + 64 = 145

after status:
pending: 388
under discussion: 10
approved: 50
not approved: 2103
sent: 8990
invalid: 207

````

jsnshrmn:
I'm more closely comparing the future science apps before and after to see if anything jumps off the screen.

jsnshrmn:
These were all applications without a stream specified, which makes them pretty questionable IMO, and I'm guessing that they get cascade deleted when the partners do. The lack of stream id would explain why authorizations aren't impacted. We could probably preserve these if we need to for statistics, but I can't think of another reason we would need to keep them. @Samwalton9, what do you think?

jsnshrmn (approved):

jsnshrmn:
I had a back-channel conversation with @Samwalton9 who gave us the go ahead.

jsnshrmn (approved):
Thanks for working this out!

PR #805, opened 2021-09-05
fix horizontal scroll bar

akayjoshi:
## Description
In My Library, there is a small amount of white space that causes a horizontal scroll bar. it should not display that horizontal scroll bar so the library fits the screen 100%.

## Phabricator Ticket
  https://phabricator.wikimedia.org/T289256

## Types of changes

### DIFF
--- a/TWLight/static/css/new-local.rtl.css
+++ b/TWLight/static/css/new-local.rtl.css
@@ -21,6 +21,10 @@ body {
   text-align: right;
 }
 
+.overflow-hidden{
+    overflow: hidden;
+}
+
 /* Different buffers for the footer based on screen size */
 @media (max-width: 767px) {
   body {
--- a/TWLight/templates/new_base.html
+++ b/TWLight/templates/new_base.html
@@ -54,7 +54,7 @@
   {% else %}
     <body>
   {% endif %}
-      <div>
+      <div class="overflow-hidden">
         {% block content %}
         {% endblock %}
       </div>


suecarmol (changes requested):
Thank you for fixing this. Users that use The Wikipedia Library on their phones will appreciate it! One small addition you need to make for this PR is to add the `overflow-hidden` class to the `new-local-rtl.css` file: https://github.com/WikipediaLibrary/TWLight/blob/master/TWLight/static/css/new-local.rtl.css#L23. This class should only have `overflow: hidden` attribute.

akayjoshi:
Hi suecarmol, I have a quetion.  Do we need to add  ```overflow-hidden``` class  in /static/css/new-local.css  also.

suecarmol:
> Hi suecarmol, I have a quetion. Do we need to add `overflow-hidden` class in /static/css/new-local.css also.

No, I don't think that's necessary because the `overflow-hidden` exists in the bootstrap 4 css. It's funny that this class doesn't exist in the RTL bootstrap 4 version.

akayjoshi:
Hi suecarmol, I have pushed a new commit for this, thanks.

suecarmol (approved):
I have checked in RTL and LTR languages, and the horizontal scroll bar bug is fixed. Thank you for your contribution!

akayjoshi:
Thank you 😊.

PR #815, opened 2021-09-19
set date_closed while withdrawing an application

amrit-srivastava:

## Description
When a user withdraws an application it should set the `date_closed` property on it. So that `get_num_days_open` gets calculated properly and we don't see a server error on the detail page of the application.

## Phabricator Ticket
https://phabricator.wikimedia.org/T269730

## How Has This Been Tested?
I have added a new test case for this behavior.

## Types of changes

### DIFF
--- a/TWLight/users/tests.py
+++ b/TWLight/users/tests.py
@@ -335,6 +335,8 @@ class ViewsTestCase(TestCase):
             request, pk=self.editor1.pk, id=app.pk
         )
         app.refresh_from_db()
+        # withdrawing application should set date closed
+        self.assertNotEqual(app.date_closed, None)
         self.assertEqual(app.status, Application.INVALID)
 
     def test_sent_application(self):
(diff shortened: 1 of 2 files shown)


TWLight/users/tests.py
```diff
@@ -337,6 +337,25 @@ def test_withdraw_application(self):
         app.refresh_from_db()
         self.assertEqual(app.status, Application.INVALID)
 
+    def test_withdraw_application_date_closed(self):
```
suecarmol, line 340: Since the difference between this test and the one above is only the `date_closed` assertion, I suggest you add that assertion in the previous test, with a comment that this assertion is included after T269730 bug fix.

suecarmol (changes requested):
Thank you for working on this task. This needs some additional work before we can merge this.

TWLight/users/tests.py
amrit-srivastava: ok @suecarmol, I will edit that and will update you soon. :)

amrit-srivastava: Hi I have made changes according to your suggestion, thanks :)

suecarmol (approved):
Thank you for fixing this bug!

PR #820, opened 2021-09-21
Improve workflow for suggestions from non-logged in users(T251991)

himanshukadian:
Improve workflow for suggestions from non-logged in users(T251991)

## Description
Improve workflow for suggestions from non-logged in users(T251991)

## Rationale
Users can suggest we pursue a new partnership at https://wikipedialibrary.wmflabs.org/suggest/. This works as expected when the user is logged in.

When the user is logged out, however, they can still enter data to the form. Clicking Submit will attempt to log them in, but then their previously written content is lost.

If this isn't an easy fix, we could instead not allow users to submit the form, prompting them to log in before they write anything.

## Phabricator Ticket
https://phabricator.wikimedia.org/T251991

## How Has This Been Tested?
Yes, manually.

image

## Types of changes
- [] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [] Minor change (fix a typo, add a translation tag, add section to README, etc.)

### DIFF
--- a/TWLight/resources/templates/resources/suggest.html
+++ b/TWLight/resources/templates/resources/suggest.html
@@ -115,4 +115,14 @@
   <script>
     function viewUpvoters(e,suggestion,box){}
   </script>
+  {% if not user.is_authenticated %}
+    <script>
+      {% comment %}Translators: This is to show alert to login first if unauthenticated user tries to fill suggestion form and then will be redirected to login page. {% endcomment %} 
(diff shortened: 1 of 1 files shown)


jsnshrmn:
Thanks for your work on this! We talked this over in engineering today, and we don't think this approach is a good fit for the platform. Really the issue is how we're defining the problem and the broadness of the acceptance criteria. I'm going to close this PR out, and we'll have more discussion about it on the phabricator task. Please know that this is not a judgement on the quality of your PR, we've just realized we need to look at the problem more closely.

PR #829, opened 2021-09-26
non-eligible users should not able to apply

amrit-srivastava:

## Description
When the non-eligible users try to apply from the /partners page then there Apply button is disabled and if they try to fill an application by directly navigating to an application page then it redirects users to My Library, where they will see the eligibility popup.

## Phabricator Ticket
https://phabricator.wikimedia.org/T290886

## How Has This Been Tested?
 I have tested this locally and also by 'python manage.py test' command.

## Types of changes


TWLight/applications/views.py
```diff
@@ -474,6 +474,20 @@ def dispatch(self, request, *args, **kwargs):
             messages.add_message(request, messages.ERROR, message)
             return HttpResponseRedirect(url, message)
 
+        # Non-elligible users should redirect to my library page
+        editor = self.request.user.editor
+        is_editor_elligible = all(
```
jsnshrmn, line 479: Eligibility is actually a little more complicated than this, which is why we have a helper that you can import:
`from TWLight.users.helpers.editor_data import editor_bundle_eligible`
It takes an editor object as the argument and returns true/false. This whole block of code should be able to be replaced with something like:
```
        if not editor_bundle_eligible(self.request.user.editor):
            url = reverse("users:my_library")
            return HttpResponseRedirect(url)
```

TWLight/resources/views.py
```diff
@@ -175,20 +175,30 @@ def get_context_data(self, **kwargs):
         context["apply"] = False
         context["has_open_apps"] = False
         context["has_auths"] = False
+        user = self.request.user
```
jsnshrmn, line 178: same here; this can be streamlined (and more correct) by using the existing helper

jsnshrmn (changes requested):
Thanks for your work on this! it does needs some more work before it can be merged in.
General comments:
- please don't create convenience variables if they are only used once (eg. `editor = self.request.user.editor`)
- it's `eligible` rather than `elligible`
- it's always best to check the source code to see if a problem has already been solved. Checking user eligibility is a thing we need to do in many places, so we have a helper for that.

More detailed feedback inline.

This is a great start!

amrit-srivastava:
I have edited some tests since `EditorFactory` always creating a `non-eligible` editor and with that when we call the `SubmitSingleApplicationView` it redirects it to the `my_library` page without creating an application according to my new changes.
I could not found a way to create an eligible editor from the `EditorFactory` it sets back to non-eligible due to some checks in models.py.

jsnshrmn (changes requested):
Thanks for your work on this; it's getting closer! The code that's here is looking great, with one inline nitpick.
I pulled this down and manually tested it and found that it does work on the single partner application button as expected.
~~However, the `RequestApplicationView` (Accessible at `/applications/request/`, which is linked at the top of `/partners/`) still allows ineligible users to apply, so it (and its tests) need the same kind of changes that you did for the individual partner view.~~
I just realized the multiple partner application case has its own separate task. Ignore that bit.

PR #830, opened 2021-09-27
user language preference update

akayjoshi:
## Description
Unauthenticated users can't change the language before login, they should be able to change language without login that takes effect for the user's session.
When logged-in users select a language that differs from their browser language we should respect that and always update the interface to display in the language selected in their preferences.
When a user logs in with a different device or incognito mode the website language should be set to the user's language preference.

## Phabricator Ticket
https://phabricator.wikimedia.org/T226804
https://phabricator.wikimedia.org/T291474

changelang

## Types of changes

### DIFF
--- a/TWLight/forms.py
+++ b/TWLight/forms.py
@@ -1,0 +1,13 @@
+from django.conf import settings
+from django import forms
+from crispy_forms.helper import FormHelper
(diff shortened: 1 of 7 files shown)


jsnshrmn:
Closing since we need a different approach.

PR #866, opened 2021-11-02
Avoid meta as first login

suecarmol:

## Description
Redirect new users to the Homepage when they click on a proxied link and they are not authenticated.

## Rationale
Current behavior causes confusion.

## Phabricator Ticket
[T292509](https://phabricator.wikimedia.org/T292509)

## How Has This Been Tested?

## Types of changes

### DIFF
--- a/TWLight/users/oauth.py
+++ b/TWLight/users/oauth.py
@@ -1,4 +1,5 @@
 import logging
+import base64
 from mwoauth import ConsumerToken, Handshaker, AccessToken
 from mwoauth.errors import OAuthException
 import urllib.parse
@@ -17,6 +18,7 @@ from django.utils.translation import get_language, gettext as _
 from urllib.parse import urlencode
 
 from .models import Editor
+from .views import OauthPromptView
 
 
 logger = logging.getLogger(__name__)
@@ -260,6 +262,10 @@ class OAuthInitializeView(View):
         # The Sites framework was designed for different URLs that correspond to
         # different databases or functionality - it's not a good fit here.
         domain = self.request.get_host()
+
+        # preference for being prompted before leaving site to authenticated.
+        oauth_prompt = request.session.get("oauth_prompt", True)
+
         try:
             assert domain in settings.ALLOWED_HOSTS  # safety first!
         except (AssertionError, DisallowedHost) as e:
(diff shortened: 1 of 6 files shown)


TWLight/users/oauth.py
```diff
@@ -324,10 +327,31 @@ def get(self, request, *args, **kwargs):
...
 
-            local_redirect = _localize_oauth_redirect(redirect)
+            # Create a QueryDict from the 'get' session dict.
+            query_dict = QueryDict(urlencode(request.session["get"]), mutable=True)
+            # Pop the 'next' parameter out of the QueryDict.
+            next = query_dict.pop("next")
+            # Set the return url to the value of 'next'. Basic.
+            return_url = next[0]
+
+            if "oclc" in return_url or "ezproxy" in return_url:
```
Samwalton9, line 337: Quick question on this - is there a particular reason to limit this behaviour to proxy links? I'm also thinking about other login-required pages, for example if someone links directly to /my_library, where this would be helpful.

suecarmol: Solved!

jsnshrmn (changes requested):
This is pretty elegant!
A few thoughts:
- let's push this to staging to make sure it the alert message doesn't goof with anonymously cached pages
- you have a few dangling imports, I'm guessing from my initial exploratory work
  - `base64`
  - `Hidden`
- Non-blocking thought: Since we're already parsing the request object, it seems like we could check `request.META.HTTP_REFERER` instead of adding a query parameter. That should always be set when navigating within our site, so you could check specifically for the root of the platform. This has a very tiny advantage of not being able to get goofed up if a stray proxy link somehow got `from_homepage=True`. That seems very unlikely, and honestly this is more of a vague aesthetic preference on my part.

suecarmol:
>     * Non-blocking thought: Since we're already parsing the request object, it seems like we could check `request.META.HTTP_REFERER` instead of adding a query parameter. That should always be set when navigating within our site, so you could check specifically for the root of the platform. This has a very tiny advantage of not being able to get goofed up if a stray proxy link somehow got `from_homepage=True`. That seems very unlikely, and honestly this is more of a vague aesthetic preference on my part.

I tried getting the `request.META.HTTP_REFERRER` value but it was always blank. I think it is because the path is external from TWL

jsnshrmn (approved):

jsnshrmn:
> let's push this to staging to make sure it the alert message doesn't goof with anonymously cached pages
oopsie, I merged anyhow. If this does cause an issue, it will be a very small fix in nginx, and would only impact the appearance (or lack thereof) of that warning message.

jsnshrmn:
Anonymous cache looks alright. The message displayed as expected when clicking a login_required link and the anonymous page cache wasn't polluted with the alert.

PR #871, opened 2021-11-04
Create a URL which initiates a library search

jsnshrmn:

## Description
Adds:
- an intentionally minimal page with  an intermediary form at `/search/?q=<search terms>` that sends users from stable urls on-wiki to searches in EDS.
- an eligible editors only view mixin that raises permission denied for users without an editor and redirects ineligible editors to my_library, which explains eligibility

Updates:
- the referring url policy to be in django instead of nginx. the header was getting added twice and it was causing intermittent problems during testing
- Drops the `from_homepage` query parameter once our auth check is done using it

## Rationale
For use in the Find Sources template

## Phabricator Ticket
https://phabricator.wikimedia.org/T294919

## How Has This Been Tested?
I added unit tests for the new view mixin, since that is reusable and has security implications.
I manually tested:
- multiple locales, including those that need the code manipulated, like `pt`, with JavaScript enabled and disabled.
- eligible and ineligible editor accounts
- anonymous and logged in users

nojs screenshot; the javascript version automatically sends the user to EDS.
image

## Types of changes


TWLight/urls.py
```diff
@@ -88,6 +88,11 @@
     url(r"^contact/$", ContactUsView.as_view(), name="contact"),
     url(r"^$", NewHomePageView.as_view(), name="homepage"),
     url(r"^about/$", TemplateView.as_view(template_name="about.html"), name="about"),
+    url(
+        r"^search$",
```
suecarmol, line 92: Did you leave the `/` out on purpose? Otherwise, I think you should add it

suecarmol (changes requested):
Just a small change before I can merge this

TWLight/urls.py
jsnshrmn: Yeah, it really is out of alignment with literally every other route. Fixing!

jsnshrmn: Done!

suecarmol (approved):
Thank you for working on this!

PR #882, opened 2021-11-10
Update My Applications to Bootstrap 4

suecarmol:

## Description
Migrated the  `My Applications` template  to use Bootstrap 4

## Rationale
Bootstrap 3 is EOL

## Phabricator Ticket
[T293988](https://phabricator.wikimedia.org/T293988)

## How Has This Been Tested?
Tested manually

Screen Shot 2021-11-09 at 19 21 31

## Types of changes

### DIFF
--- a/TWLight/static/css/new-local.css
+++ b/TWLight/static/css/new-local.css
@@ -405,6 +405,16 @@ COMMON CONTENT BLOCK CSS
     text-decoration: underline;
 }
 
+.twl-secondary-btn{
+    color: #000000;
+    background-color: #FFFFFF;
+    border-color: #CCCCCC;
+}
+
+.twl-secondary-btn:hover{
+    background-color: #CCCCCC;
+}
+
 /*
 ----------------------------------------------------------------------------
 NOSCRIPT CSS
--- a/TWLight/static/css/new-local.css
+++ b/TWLight/static/css/new-local.css
@@ -1254,7 +1254,7 @@ CONTACT US CSS
 --------------------------------------------------------------------------
 */
 
-.contact-us-links{
+.contact-us-links, .evaluate-application-link{
     color: #3366CC;
 }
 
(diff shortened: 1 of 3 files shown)


TWLight/users/templates/users/my_applications.html
```diff
@@ -1,91 +1,95 @@
...
-        <h4>
-        <a href="{{ app.get_absolute_url }}">
-          {{ app.partner }}
-        </a>
-        </h4>
-        <div class="visible-xs">
-          <span class="label label{{ app.get_bootstrap_class }}">
+  {% include "header_partial_b4.html" %}
+  <div id="main-content">
+    {% include "message_partial.html" %}
```
jsnshrmn, line 8: nitpick: This partial is outside the main-content div in all other cases. Having it here doesn't break anything on the page, but it does result in the messages having a different position than all the other pages that use the partial.

jsnshrmn (changes requested):
I have one inline nitpick, but this is otherwise ready to merge

jsnshrmn (approved):

TWLight/static/css/new-local.css
```diff
@@ -405,6 +405,16 @@ COMMON CONTENT BLOCK CSS
     text-decoration: underline;
 }
 
+.twl-secondary-btn{
```
jsnshrmn, line 408: having just gone and cleaned up my own rtl omissions, I just realized that these also need to be added to `new-local.rtl.css`

jsnshrmn (changes requested):
I realized that we overlooked the RTL css.
I've got more cleanup in my outstanding pr, but let's just tackle these new classes in this one.

TWLight/static/css/new-local.css
suecarmol: Good catch! Thanks for spotting that 😄

jsnshrmn (approved):

PR #923, opened 2021-12-21
Add wp_block_hash field

suecarmol:

## Description
Added the `wp_block_hash` field and validated if the block has has changed everytime the user logs in

## Rationale
We want a system for reviewing users with ignored block status in the case that their block status changes, so that we can evaluate if they should still be added to the allow list.

## Phabricator Ticket
[T295608](https://phabricator.wikimedia.org/T295608)

## How Has This Been Tested?

## Types of changes


jsnshrmn (approved):
Thanks for all of your work on this! It is looking really good. There is only one thing that sticks out at me, and it's not actually a problem. It's the fact that the email send is setup directly in the user helper. We're using signals for sending emails in other places, but I don't think there is any advantage to registering a signal here; it has me wondering why I implemented so many of the other signals. Great work!

PR #968, opened 2022-02-25
Don't send renewal reminder email when a renewal has already been filed

suecarmol:

## Description
Don't send a renewal reminder email when a renewal has already been filed.

## Rationale
Two weeks before an authorization is due to expire we send users an email notifying them of this and suggesting they request a renewal if they want to continue having access. This is confusing for users who anticipated this and already filed a renewal request.

## Phabricator Ticket
[T300014](https://phabricator.wikimedia.org/T300014)

## How Has This Been Tested?
Added a new test

## Types of changes

### DIFF
--- a/TWLight/users/management/commands/user_renewal_notice.py
+++ b/TWLight/users/management/commands/user_renewal_notice.py
@@ -1,6 +1,8 @@
 from datetime import datetime, timedelta
 
+from django.contrib.auth.models import User
 from django.core.management.base import BaseCommand
+from django.db.models import Prefetch
 from django.urls import reverse
 
 from TWLight.users.signals import Notice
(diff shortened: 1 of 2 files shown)


TWLight/users/management/commands/user_renewal_notice.py
```diff
@@ -13,13 +17,32 @@ class Command(BaseCommand):
...
-        expiring_authorizations = Authorization.objects.filter(
-            date_expires__lt=datetime.today() + timedelta(weeks=2),
-            date_expires__gte=datetime.today(),
-            reminder_email_sent=False,
-            partners__isnull=False,
-        ).exclude(user__userprofile__send_renewal_notices=False)
+        # exclude users who disabled these emails and who have already filed
+        # for a renewal.
+        editor_qs = Editor.objects.select_related("user")
+        users_with_applications_for_renewal = (
```
Samwalton9, line 23: Just quickly looking at this so sorry if I've missed something - we specifically want to discount users who have a renewal request for the resource to which their authorization is expiring. Is this queryset editors with any renewal request or specifically for this resource?

suecarmol: Great question. I will create a test case where a user has 2 soon-to-expire auths and has just renewed one. They should only get one email. That will test if this code is working correctly :)

suecarmol: You were absolutely right: the code as it was didn't work as expected. Have changed it and added new test cases.

jsnshrmn (approved):
:+1: thanks for your work; it's looking good to me!

PR #983, opened 2022-03-26
Only send waitlist email once per application

Samwalton9:
_Submitted in my capacity as a volunteer :)_

## Description
I added a new field to the application model to track whether we sent a waitlist email for that application, set it to True when the email is sent, and checked for it in the relevant email task.

## Rationale
We shouldn't send users multiple waitlist emails if a partner gets switched back and forth between waitlisted and available, as happens for busy proxy partners which have authorizations expiring on a regular basis.

## Phabricator Ticket
https://phabricator.wikimedia.org/T304512

## How Has This Been Tested?
I added a test, which was failing before I made the code changes (2 emails sent instead of the desired 1).

## Types of changes
- [X] New feature (non-breaking change which adds functionality)

### DIFF
--- a/TWLight/emails/tasks.py
+++ b/TWLight/emails/tasks.py
@@ -350,6 +350,8 @@ def send_waitlist_notification_email(instance):
                     "link": link,
                 },
             )
+            instance.waitlist_email = True
+            instance.save()
         else:
             logger.info(
                 "Skipped user {username} when sending "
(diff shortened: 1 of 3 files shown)


TWLight/emails/tasks.py
```diff
@@ -481,19 +483,19 @@ def send_authorization_emails(sender, instance, **kwargs):
...
+    who hasn't already received an email for this application should be notified.
     """
     if instance.id:
         orig_partner = get_object_or_404(Partner, pk=instance.id)
 
-        if (
-            orig_partner.status != instance.status
-        ) and instance.status == Partner.WAITLIST:
-
+        status_changed = orig_partner.status != instance.status
```
Samwalton9, line 491: This change has nothing directly to do with the task, I just wanted to tidy up the superfluous brackets and clarify what was happening.

suecarmol:
Hi, volunteer Sam :) Can you please pull the latest changes from the `master` branch and rebase this PR?

suecarmol:
Looks like the build is erroring out on a test.

Samwalton9:
I think I see the source of the issue - we've got circular signalling because I added a `.save()` in a function called by the signal, which prompts the signal again, ad infinitum.

PR #984, opened 2022-03-27
Don't email about blocks on exempted users first login

Samwalton9:
_Submitted in my capacity as a volunteer :)_

## Description
When users login for the first time, we were always setting their block hash based on an empty dictionary. This change sets it based on their block string, which is an empty dictionary if they're not blocked, but contains their block information if they are blocked.

## Rationale
When a user is already blocked on their first login, gets exempted, and logs in a 2nd time, we currently receive a spurious email about block status changes. This should fix that.

## Phabricator Ticket
https://phabricator.wikimedia.org/T301668

## How Has This Been Tested?
I added a new test, which was failing on line 1562 before I made my change (the user's block dictionary wasn't matching what we'd saved for block hash).

## Types of changes

### DIFF
--- a/TWLight/users/helpers/editor_data.py
+++ b/TWLight/users/helpers/editor_data.py
@@ -294,5 +294,6 @@ def editor_compare_hashes(
         else:
             return previous_block_hash
     else:
-        # previous_block_hash is blank, we will make a hash of an empty dictionary
-        return make_password("{}")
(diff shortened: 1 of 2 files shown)


suecarmol:
Hi, volunteer Sam :) Can you please pull the latest changes from the `master` branch and rebase this PR?

suecarmol (approved):
Thank you for fixing this spammy bug :)

PR #988, opened 2022-04-07
Don't create duplicate Bundle authorization when non-available resources are moved to Bundle

lalit97:

## Description
When a partner is moved from an access method to Bundle we do some processing in [update_existing_bundle_authorizations](https://github.com/WikipediaLibrary/TWLight/blob/fe6369f1365ad8f53b630a009583ed3ba4811d7b/TWLight/users/signals.py#L72) to update authorizations accordingly.

When the partner is set to one of the non-available statuses and is moved to Bundle, it creates duplicate Bundle authorizations for users. This shouldn't happen.

## Rationale
When the partner is set to one of the non-available statuses and is moved to Bundle in the same operation, 
it creates duplicate Bundle authorizations for the user. This should not happen ideally.

## Phabricator Ticket
https://phabricator.wikimedia.org/T302882

## How Has This Been Tested?
I have tested it by moving a partner from `available` to `waitlist` and setting its `authorization_method` to `Bundle` in the same query. Later checked that it does not have duplicate Authorizations in the DB.

## Types of changes

### DIFF
--- a/TWLight/users/signals.py
+++ b/TWLight/users/signals.py
@@ -107,12 +107,11 @@ def update_existing_bundle_authorizations(sender, instance, **kwargs):
         if previously_available and previously_bundle:
             remove_from_auths = True
 
-        # Case when partner is no more available
-        # but set to Bundle at the same time.
-        # We need to delete its existing authorizations for now
-        # whenever its status or authorization_method will change
-        # in future, situation will be handled by either add_to_auths
-        # remove_from_auths accordingly
+        # Fix T302882: Case when a partner does not have an Available status
+        # and has a Bundle authorization method set in the same operation.
+        # Then we need to delete its existing authorizations for now.
+        # If its status or authorization_method changes in future, situation will be
+        # handled by either add_to_auths or remove_from_auths accordingly.
         elif now_bundle and not previously_bundle:
             delete_defunct_authorizations = True
 
--- a/TWLight/users/signals.py
+++ b/TWLight/users/signals.py
@@ -78,6 +78,7 @@ def update_existing_bundle_authorizations(sender, instance, **kwargs):
     """
     add_to_auths = False
     remove_from_auths = False
+    delete_defunct_authorizations = False
 
     try:
         previous_data = Partner.even_not_available.get(pk=instance.pk)
@@ -106,8 +107,17 @@ def update_existing_bundle_authorizations(sender, instance, **kwargs):
         if previously_available and previously_bundle:
             remove_from_auths = True
 
+        # Case when partner is no more available
+        # but set to Bundle at the same time.
+        # We need to delete its existing authorizations for now
+        # whenever its status or authorization_method will change
+        # in future, situation will be handled by either add_to_auths
+        # remove_from_auths accordingly
+        elif now_bundle and not previously_bundle:
+            delete_defunct_authorizations = True
+
     # Let's avoid db queries if we don't need them
-    if add_to_auths or remove_from_auths:
+    if add_to_auths or remove_from_auths or delete_defunct_authorizations:
         authorizations_to_update = get_all_bundle_authorizations()
 
         if add_to_auths:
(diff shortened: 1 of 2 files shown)


suecarmol:
Hi @lalit97! We are having some trouble trying to reproduce this error. Could you please add detailed instructions about how you reproduced this bug and the queries you made to check that

1. the duplicate authorizations were created
2. the bug was fixed when you made your changes.

suecarmol:
> Hi @lalit97! We are having some trouble trying to reproduce this error. Could you please add detailed instructions about how you reproduced this bug and the queries you made to check that
> 
>     1. the duplicate authorizations were created
> 
>     2. the bug was fixed when you made your changes.

I was able to reproduce this bug and added the steps to reproduce it to the Phab task, although it will still be good to know how you reproduced this error in your own words :)

TWLight/users/signals.py
```diff
@@ -106,8 +107,17 @@ def update_existing_bundle_authorizations(sender, instance, **kwargs):
         if previously_available and previously_bundle:
             remove_from_auths = True
 
+        # Case when partner is no more available
+        # but set to Bundle at the same time.
```
suecarmol, line 111: Let's add the Phab ticket at the beginning of this comment. Also, let's change the wording a bit from `Case when partner is no more available` to `Case when a partner does not have an Available status and has a Bundle authorization method set in the same operation.`

TWLight/users/signals.py
```diff
@@ -106,8 +107,17 @@ def update_existing_bundle_authorizations(sender, instance, **kwargs):
...
+        # We need to delete its existing authorizations for now
+        # whenever its status or authorization_method will change
```
suecarmol, line 113: Start a new sentence here: `If its status or authorization_method changes...`

suecarmol (commented):
Thank you for working on this tricky bug fix. I have some initial minor comments,

TWLight/users/signals.py
lalit97: Thank you for the suggestions :) I have updated the comment

lalit97:
> Hi @lalit97! We are having some trouble trying to reproduce this error. Could you please add detailed instructions about how you reproduced this bug and the queries you made to check that
> 
> 1. the duplicate authorizations were created
> 2. the bug was fixed when you made your changes.

Hi @suecarmol  sure, I will add the steps here soon :)

lalit97:
> 1. the duplicate authorizations were created

First of all, I selected a Bundle eligible user (having wp_bundle_eligible=True). So the user was `Authorized` for all the partners who had the `Bundle` authorization_method.

```
In [1]: from TWLight.users.models import Authorization

In [2]: from django.contrib.auth.models import User

In [3]: user = User.objects.get(username="61329345")

In [4]: for auth in Authorization.objects.filter(user=user):
                 print(auth)
Out [4]:
authorized: Lalit97meta - authorizer: TWL Team - date_authorized: 2022-05-03 - company_name: Freudenberger, Geisler, Pohl, Sager KG, Salz, Zobel GbR

```
Then I created an `application` with that user for a non-Bundle partner (Austermühle) and approved it. So now there was a new `Authorization` for the user for that given partner.

```
In [5]: for auth in Authorization.objects.filter(user=user):
                 print(auth)
Out [5]: 
authorized: Lalit97meta - authorizer: TWL Team - date_authorized: 2022-05-03 - company_name: Freudenberger, Geisler, Pohl, Sager KG, Salz, Zobel GbR
authorized: Lalit97meta - authorizer: lalit_staff - date_authorized: 2022-05-03 - company_name: Austermühle

```
Then I changed the partner's `status` and `authorization_method` in a single operation like:

```
In [7]: from TWLight.resources.models import Partner
In [8]: partner = Partner.objects.get(id=48)

In [10]: partner.authorization_method == Partner.EMAIL
Out[10]: True

In [12]: partner.status == Partner.AVAILABLE
Out[12]: True

In [14]: partner.authorization_method = Partner.BUNDLE
In [15]: partner.status = Partner.WAITLIST
In [16]: partner.save()

```
I tried to print all the `Authorizations` for the `user` and got these results: 

```
In [17]: for auth in Authorization.objects.filter(user=user):
                   print(auth)
Out [17]:
authorized: Lalit97meta - authorizer: TWL Team - date_authorized: 2022-05-03 - company_name: Freudenberger, Geisler, Pohl, Sager KG, Salz, Zobel GbR
authorized: Lalit97meta - authorizer: lalit_staff - date_authorized: 2022-05-03 - company_name: Austermühle

```
Now the Austermühle partner moved to `Bundle`, so only 1 `Bundle` type `Authorization` should be present for each `user`. But I still received 2 `Authorization` objects for the `user`. Which meant `Bundle authorization` got duplicated.

lalit97:
No issues at all. Thank you :grinning:

PR #997, opened 2022-04-22
WIP - Add method to check for user preferred language on login

suecarmol:

## Description
- Added a new method to check if a user's preferred language is the same as the browser language
   - If they are different, then we set the language to the user's preferred language
   - If they haven't set a language, it will use the browser language

## Rationale
When users select a language that differs from their browser language we should respect that and always update the interface to display in the language selected in their preferences.

## Phabricator Ticket
[T291474](https://phabricator.wikimedia.org/T291474)

## How Has This Been Tested?

## Types of changes


jsnshrmn (approved):
Thanks for the fix! This looks great!

PR #1065, opened 2022-10-04
Add PhabricatorTask class to relate phab tasks to partner collections

jsnshrmn:

## Description
- resources: adds PhabricatorTask class to relate phab tasks to partner collections
- users: updates view and templates to display task buttons alongside access button

## Rationale
When a collection has a short-term issue, we want to link users to a phab ticket with more info

## Phabricator Ticket
https://phabricator.wikimedia.org/T314780

## How Has This Been Tested?
I just manually tested this by clicking around in my librarry

image

## Types of changes


TWLight/resources/models.py
```diff
@@ -469,6 +482,50 @@ class PartnerLogo(models.Model):
...
+        blank=True,
+        # Limit to available partners.
+        help_text="The partner(s) affected by this task.",
+    )
+
+    INFORMATION = 0
+    WARNING = 1
+    OUTAGE = 2
+
+    TYPE_CHOICES = (
```
Samwalton9, line 501: This is a really nice approach, thanks for thinking beyond our initial scope!

jsnshrmn:
:point_up: just realized I hadn't pushed a small amendment

jsnshrmn:
>     1. Should we disable the `Access collection/Apply` button when there is a service outage?

Yes for access and no for apply, IMO. Applications can take some time, and the outage might be resolved by the time they are approved. I was thinking that replacing the access button might be even better than replacing it. I kind of feel like we should pull some logic out of the template and into the view so that the template is just rendering arbitrary buttons in the card footer. Rather than shaving that yak, I stuck this in review. Probably the right thing to do is just add a *little more* conditional logic to the template to disable/replace the access button to get the easy win.
 
>     2. UI nitpick: The new buttons do not have the same hover behavior as the `Access collection/Apply` button.
Noted!
>     3. UI nitpick: I think a better choice of button type is the [outline buttons](https://getbootstrap.com/docs/4.0/components/buttons/#outline-buttons). Having this be a regular button gives it the same relevance as the `Access collection/Apply` button. I think this button can have the same relevance as the `Renew/Extend` button which is an outline button.
💯 
>     4. Should we add an explanation as to what each status means? Maybe by adding a [collapse](https://getbootstrap.com/docs/4.0/components/collapse/#example) and then linking to the phab task there?
Thanks for the comments on this! I treated these like simple "call to action" buttons because that was the shortest path to show that there is an issue and link to the related task.  Those really are two separate things though. I'll take another pass at it.

jsnshrmn:
here's a somewhat naive approach:
image
I changed the text to all-black to maintain good color contrast. The expanded div isn't super graceful looking (This was low effort), but I think this is much better than what I originally had.

suecarmol (approved):
This looks good! I will wait until @Samwalton9's approval to merge this.

Samwalton9:
I think this is great, go ahead!

jsnshrmn:
I just realized that this didn't have rtl styling because we were using a manually created rtl stylesheet. I went ahead and dropped the manually created stylesheet and updated the reference in the template. I did manual local testing in hebrew (which requires running bin/twlight_cssjanus.js) and the results are looking good. Also pushing this to staging, though it will take a bit to deploy.

suecarmol:
LGTM!

PR #1090, opened 2022-11-24
Improve Wikilala's English description

amire80:
"Existence" doesn't sound right in English. I discussed it with the Turkic Wikimedians group, and they agreed that "founding" makes for sense.


suecarmol (approved):
Thank you for submitting this correction!

PR #1115, opened 2023-01-19
users favorites: only validate favorites being added; not the entire set of favorites

jsnshrmn:

## Description
only validates partners being favorited upon change; If a user does somehow manage to attempt to favorite an unauthorized partner, the error is now visible to them. Also makes favorites visible in user admin for troubleshooting.

## Rationale
currently we validate the entire set of favorites, which breaks favoriting for users unnecessarily; the rest endpoint for favoriting was responding with our themed error page instead of a json response, so the error couldn't be passed back to the user via ajax.

## Phabricator Ticket
https://phabricator.wikimedia.org/T323216

## How Has This Been Tested?
I identified the issue by pulling a backup and overwriting the wiki references for an impacted user with my own wiki id. I then manually tested that the fix resolved the problem. It also passes our unit tests

image

## Types of changes

### DIFF
--- a/TWLight/users/admin.py
+++ b/TWLight/users/admin.py
@@ -14,6 +14,8 @@ from TWLight.users.models import (
 )
 from TWLight.users.forms import AuthorizationAdminForm, AuthorizationInlineForm
 
+from TWLight.resources.models import Partner
+
 
 class EditorInline(admin.StackedInline):
     model = Editor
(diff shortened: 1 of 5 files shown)


jsnshrmn:
I jumped the gun: I need to update the test!

suecarmol (approved):
Thank you for working on this, Jason! I've tested this locally, and favoriting collections are now working as expected :)

PR #1163, opened 2023-05-19
Increase character limit on suggest page

NovemLinguae:
From 40 characters to 80 characters

Bug: T330457


suecarmol (approved):
LGTM! Thank you for working on this patch!

PR #1173, opened 2023-06-23
WIP dependencies: drop pytz

jsnshrmn:

## Description
This just drops the pytz dependency. It looks like this was only needed for older versions of django.

## Rationale

## Phabricator Ticket
N/A

## How Has This Been Tested?

## Types of changes


jsnshrmn:
From my reading of the docs: while it looks like django 3.2 supports the built in zoneinfo of newer python builds, it's still going to silently install pytz as a dependency. We can just save this for django 4.x, which drops pytz altogether.

PR #1192, opened 2023-08-09
T254975: Improve Library Card embed

bharatswamy:

## Description
Added open graph meta tags to beautify the Wikipedia library embed link 

## Rationale
When a link to the Library Card platform homepage is embedded on another website, such as Facebook or Slack, the result isn't pretty,

## Phabricator Ticket
https://phabricator.wikimedia.org/T254975

## How Has This Been Tested?
Tested by deploying to the local server and pasting the link to open graph debugger by Facebook
https://developers.facebook.com/tools/debug/

image

## Types of changes

### DIFF
--- a/TWLight/templates/new_base.html
+++ b/TWLight/templates/new_base.html
@@ -16,11 +16,20 @@
     <meta property="og:image:width" content="200" />
     <meta property="og:image:height" content="200" />
     <meta property="og:url" content="https://wikipedialibrary.wmflabs.org/" />
-    <meta property="og:description" content="Over 90 of the world's top subscription-only databases, with content in 33 languages, free for Wikipedians of all backgrounds." />
+    <meta property="og:description" content="{% comment %}Translators: Information about The Wikipedia Library.{% endcomment %}
+    {% blocktranslate trimmed %}
+    Over 90 of the world's top subscription-only databases,
+    with content in {{ no_of_languages }} languages,
+    free for Wikipedians of all backgrounds
+    {% endblocktranslate %}"
+    />
     <meta property="twitter:card" content="summary_large_image" />
     <meta property="twitter:image" content="https://upload.wikimedia.org/wikipedia/commons/6/63/Wikipedia_library_logo.png" />
     <meta property="twitter:site" content="@WikiLibrary" />
-
+    <meta property="og:locale" content=en_US />
+    {% for language in all_languages %}
+      <meta property="og:locale:alternate" content={{language}} />
+    {% endfor %}
     <title>
       {% comment %}Translators: Name of the project/tool.{% endcomment %}
       {% trans "The Wikipedia Library" %}
--- a/TWLight/templates/new_base.html
+++ b/TWLight/templates/new_base.html
@@ -10,6 +10,17 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <!-- The above 3 meta tags *must* come first in the head; any other head content must come *after* these tags -->
 
+    <meta property="og:title" content="The Wikipedia Library" />
+    <meta property="og:type" content="website" />
+    <meta property="og:image" content="https://upload.wikimedia.org/wikipedia/commons/6/63/Wikipedia_library_logo.png" />
+    <meta property="og:image:width" content="200" />
+    <meta property="og:image:height" content="200" />
+    <meta property="og:url" content="https://wikipedialibrary.wmflabs.org/" />
+    <meta property="og:description" content="Over 90 of the world's top subscription-only databases, with content in 33 languages, free for Wikipedians of all backgrounds." />
+    <meta property="twitter:card" content="summary_large_image" />
+    <meta property="twitter:image" content="https://upload.wikimedia.org/wikipedia/commons/6/63/Wikipedia_library_logo.png" />
+    <meta property="twitter:site" content="@WikiLibrary" />
+
     <title>
       {% comment %}Translators: Name of the project/tool.{% endcomment %}
       {% trans "The Wikipedia Library" %}
--- a/TWLight/templates/new_base.html
+++ b/TWLight/templates/new_base.html
@@ -10,7 +10,8 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <!-- The above 3 meta tags *must* come first in the head; any other head content must come *after* these tags -->
 
-    <meta property="og:title" content="The Wikipedia Library" />
+    {% comment %}Translators: Name of the project/tool.{% endcomment %}
+    <meta property="og:title" content="{% trans "The Wikipedia Library" %}" />
     <meta property="og:type" content="website" />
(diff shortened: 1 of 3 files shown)


TWLight/templates/new_base.html
```diff
@@ -10,6 +10,17 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <!-- The above 3 meta tags *must* come first in the head; any other head content must come *after* these tags -->
 
+    <meta property="og:title" content="The Wikipedia Library" />
+    <meta property="og:type" content="website" />
+    <meta property="og:image" content="https://upload.wikimedia.org/wikipedia/commons/6/63/Wikipedia_library_logo.png" />
+    <meta property="og:image:width" content="200" />
+    <meta property="og:image:height" content="200" />
+    <meta property="og:url" content="https://wikipedialibrary.wmflabs.org/" />
+    <meta property="og:description" content="Over 90 of the world's top subscription-only databases, with content in 33 languages, free for Wikipedians of all backgrounds." />
```
jsnshrmn, line 19: In `TWLight/templates/login_partial.html`, you can see how we template a source message with the correct arguments here:
```
{% comment %}Translators: Information about The Wikipedia Library.{% endcomment %}
{% blocktranslate trimmed %}
Over 90 of the world's top subscription-only databases,
with content in {{ no_of_languages }} languages, 
free for Wikipedians of all backgrounds
{% endblocktranslate %}
```
For proper i18n, I think you need to:
- use the above message block for `og:description`
- loop through the available languages, and set each one as an `og:locale:alternate` 
- add support for setting the locale via a url parameter (which might be a big task to do correctly; I'd need to do a little research)

jsnshrmn (changes requested):
Thank you for your work in this! As we add new features, it's really important that we consider how those features will be localized, and I think there is some work to do here. This approach currently is English only, and also hardcodes the language count, which will quickly go out of date. See my inline comment for more details.

If you're wanting to dig deeper into django, i18n handling can open up lots of interesting lines of inquiry, IMO. Please let me know if you have questions or comments!

TWLight/templates/new_base.html
bharatswamy: > loop through the available languages, and set each one as an og:locale:alternate

Hi @jsnshrmn to use this we will need list of `language_TERRITORY` refer: [https://ogp.me/#locale](https://ogp.me/#array:~:text=og%3Alocale%20%2D%20The%20locale%20these%20tags%20are%20marked%20up%20in.%20Of%20the%20format%20language_TERRITORY.%20Default%20is%20en_US.)

we can get a list of languages from the `Language` model in the resources app but it does not have `TERRITORY` do we populate it somewhere in our code or should I try fetching it using some third-party package eg: https://github.com/flyingcircusio/pycountry or is there another way

jsnshrmn: That's a great question. Generally, only the language is required in XPG formatted locale names, eg. `language[_territory[.codeset]][@modifier]` where everything in brackets is optional. I would start with our locales as they are and see if that works.

Because of our integration with translatewiki, we're already crosswalking Django's locales with wikimedia CLDR data, so we could add some machinery to define default territories if necessary, but I wouldn't bother with that complexity unless it's clear that it's required for interoperability.

We do have a few locales with territories defined already that you can use for testing in case they are required:
- en_Gb
- ks-arab
- pt_Br
- sr_Ec
- zh_Hans
- zh_Hant

bharatswamy: Thank you for the precise explanation @jsnshrmn I have made changes according to your suggestion and pushed a new commit

TWLight/templates/new_base.html
```diff
@@ -10,6 +10,26 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <!-- The above 3 meta tags *must* come first in the head; any other head content must come *after* these tags -->
 
+    <meta property="og:title" content="The Wikipedia Library" />
```
jsnshrmn, line 13: I missed this before, but this should be localized as well.
```suggestion
      {% comment %}Translators: Name of the project/tool.{% endcomment %}
    <meta property="og:title" content="{% trans "The Wikipedia Library" %}" />
```

TWLight/views.py
```diff
@@ -124,6 +124,7 @@ def get_context_data(self, **kwargs):
         context["no_of_languages"] = (
             Partner.objects.values("languages").distinct().count()
         )
+        context["all_languages"] = Language.objects.values_list("language", flat=True)
```
jsnshrmn, line 127: Exclude current language here (and rename to something like alternate_languages)

TWLight/templates/new_base.html
```diff
@@ -10,6 +10,26 @@
...
+    {% blocktranslate trimmed %}
+    Over 90 of the world's top subscription-only databases,
+    with content in {{ no_of_languages }} languages,
+    free for Wikipedians of all backgrounds
+    {% endblocktranslate %}"
+    />
+    <meta property="twitter:card" content="summary_large_image" />
+    <meta property="twitter:image" content="https://upload.wikimedia.org/wikipedia/commons/6/63/Wikipedia_library_logo.png" />
+    <meta property="twitter:site" content="@WikiLibrary" />
+    <meta property="og:locale" content=en_US />
```
jsnshrmn, line 29: use the value from [django.utils.translation.get_language_from_request()](https://docs.djangoproject.com/en/3.2/ref/utils/#django.utils.translation.get_language_from_request) instead of hardcoding `en_US` Django will take care of figuring out the best locale to use.

jsnshrmn (changes requested):
Followup on my previous remarks. Thanks again for your work, this is almost there!

TWLight/templates/new_base.html
bharatswamy: Oh, I missed this as well, thank you for pointing it out. I have added the trans tag in the latest commit

TWLight/templates/new_base.html
bharatswamy: Thanks, I have removed the hardcoded `en_US` and used the current language here

TWLight/templates/new_base.html
```diff
@@ -10,6 +10,26 @@
...
+    {% for language in all_languages %}
```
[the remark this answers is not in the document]
bharatswamy: done, I have excluded the current language from this list

TWLight/templates/new_base.html
bharatswamy: Thank you for the review I have made the changes according to the suggestions using the [django.utils.translation.get_language_from_request()](https://docs.djangoproject.com/en/3.2/ref/utils/#django.utils.translation.get_language_from_request). This has really simplified the task

jsnshrmn (commented):
Sorry for the long wait on this review! The code looks good based on a read-through. I'm going to test this in our staging environment as the last step before approval as long as there are no problems.

jsnshrmn (approved):

jsnshrmn (approved):
Okay, I tested this with a few languages, and it looks good. Thanks for all of your work!

PR #1197, opened 2023-09-09
Enable users with invalid applications to delete their data

Samwalton9:

## Description
Fixed bug [T338148](https://phabricator.wikimedia.org/T338148) - users who had applications with the INVALID status were unable to delete their account because we didn't use `include_invalid` when fetching application objects to check if we should send emails for status changes.

## Phabricator Ticket
https://phabricator.wikimedia.org/T338148

## How Has This Been Tested?
Added a test, which failed until I made the code change in emails/tasks.py.

## Types of changes


suecarmol (approved):
Tested it locally and it solved the bug. Thanks for fixing this!

PR #1199, opened 2023-09-09
Add config app and CommunityPage model for local discussion pages

Samwalton9:

## Description
Per https://phabricator.wikimedia.org/T299510, we wanted to add a community page link depending on the user's language code. Non-technical WMF staff want to be able to add new community pages and associate them with a language.

I figure there were probably a few different approaches we could have taken here. The option I went for was to add a new app, `config`, and define a `CommunityPage` model there for this data. My rationale was that we might want to use the `config` app for other site-wide configuration models, which are more accessible to non-developers than requiring a config file to be updated via a pull request. I appreciate this comes with a technical overhead tradeoff, though. I'm very happy to hear feedback on whether this was a sensible approach or not and rework accordingly.

There's still work to do here, I think, namely tests. I didn't want to go too far before getting feedback on the approach.

## Rationale
Laid out in https://phabricator.wikimedia.org/T299510.

## Phabricator Ticket
https://phabricator.wikimedia.org/T299510

## How Has This Been Tested?
Local testing with languages that have and have not been defined. I didn't write any tests yet because I wanted to get feedback on the approach.

Screenshot 2023-09-09 at 16 38 05

## Types of changes

### DIFF
--- a/TWLight/users/templates/users/filter_section.html
+++ b/TWLight/users/templates/users/filter_section.html
@@ -17,13 +17,14 @@
       <div class="about-links">
         <a href="{% url 'about' %}" class="twl-links twl-links-underline learn-more-link">
           {% comment %}Translators: On the MyLibrary page (https://wikipedialibrary.wmflabs.org/users/my_library), this appears below introductory text and links to the 'About' page (https://wikipedialibrary.wmflabs.org/about/) {% endcomment %}
-          {% trans "Learn more" %}
-        </a>
+          {% trans "Learn more" %}</a>
         &ndash;
         <a href="https://meta.wikimedia.org/wiki/Talk:The_Wikipedia_Library" class="twl-links twl-links-underline discuss-link">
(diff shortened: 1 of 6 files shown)


TWLight/users/templates/users/filter_section.html
```diff
@@ -17,13 +17,14 @@
       <div class="about-links">
         <a href="{% url 'about' %}" class="twl-links twl-links-underline learn-more-link">
           {% comment %}Translators: On the MyLibrary page (https://wikipedialibrary.wmflabs.org/users/my_library), this appears below introductory text and links to the 'About' page (https://wikipedialibrary.wmflabs.org/about/) {% endcomment %}
-          {% trans "Learn more" %}
-        </a>
+          {% trans "Learn more" %}</a>
```
Samwalton9, line 20: Just a note that I moved these </a>s to prevent the underline from straying into whitespace.

Samwalton9:
An initial followup thought: Maybe a `config` app is overkill for this one `CommunityPage` model, and it should have just lived in the `users` app since it's only used within the context of that app.

suecarmol (changes requested):
Thank you for addressing this problem Sam! While your approach was sound, I think I found a better one in 1224. Let me know what you think!

Samwalton9:
Closing in favour of https://github.com/WikipediaLibrary/TWLight/pull/1224.

PR #1200, opened 2023-09-09
Fix overwriting stored email with blank string when user has no WP email

Samwalton9:

## Description
Previously, if a user did not have a Wikipedia account email, but did enter an email manually into the library, every time they logged in their email would be overwritten with a blank string. After this change, Wikipedia email data is only stored if it contains content.

## Phabricator Ticket
https://phabricator.wikimedia.org/T328904

## How Has This Been Tested?
Tested locally with a few different accounts having and not having email addresses linked. Added a test, which fails with the old code and passes with the new.

## Types of changes


jsnshrmn (approved):
the code makes sense, and local testing verifies that it works. Thanks!

PR #1202, opened 2023-09-11
Improved code documentation in resources/views.py

Samwalton9:

## Description
Ran through the entire file with fairly fresh eyes, documenting anything that was initially unclear to me.

I also refactored one piece of code where we were duplicating lines a bunch of times, adding a couple of model properties and a method instead. This could have technically been simplified further, but I tried to keep things slightly generic in case, for example, we just want to fetch a partner's short description.

## Phabricator Ticket
https://phabricator.wikimedia.org/T264424

## How Has This Been Tested?
Poked around the site locally, and ran tests.

## Types of changes
- [x] Minor change (fix a typo, add a translation tag, add section to README, etc.)


TWLight/resources/views.py
```diff
@@ -73,28 +73,32 @@ def get_context_data(self, **kwargs):
...
             self.request.GET, queryset=self.get_queryset(), language_code=language_code
         )
         context["filter"] = partner_filtered_list
 
+        # Retrieve the current user and add to context
         user = self.request.user
         if user.is_authenticated:
             user = User.objects.select_related("editor").get(pk=self.request.user.pk)
             context["user"] = user
-            context["editor"] = user.editor
```
Samwalton9, line 94: This didn't seem to be used anywhere. Replaced it with a string and couldn't find the string in page source, and couldn't find `editor` in the relevant html files. Nothing seemed to break - it looks like we're accessing the `editor` via `user.editor` in the html.

TWLight/resources/views.py
```diff
@@ -36,65 +34,70 @@
...
-                .order_by("company_name")
-            )
+            qs = Partner.even_not_available.prefetch_related("languages")
+        else:
+            qs = Partner.objects.prefetch_related("languages")
 
-        return qs
+        # The ordering here is useful primarily to people familiar with the
+        # English alphabet. :/
+        partner_queryset = qs.select_related("coordinator", "logos").order_by(
```
jsnshrmn, line 63: This is a big improvement; I think you could put the prefetch here to, for even more DRYness.

jsnshrmn (commented):
This looks good and worked as expected in my local testing. I have one non-blocking suggestion. Feel free to decline it if you don't like it! Either way, I'll happily merge.

TWLight/resources/views.py
Samwalton9: Done!

jsnshrmn (approved):
Thanks for your work on this!

PR #1209, opened 2023-09-24
Migrate single-argument helper functions to model methods

Samwalton9:

## Description
We had two functions floating around the users app which only took a single argument - a partner. It makes much more sense to me for these to be model methods. I wouldn't expect to need to go digging through helpers files to find this.

In the process of doing this I discovered one of the tangible downsides to the original setup which was the confusing `partner_pk` argument, which was sometimes an int (a pk) and sometimes a Partner object, depending on where the function was called. Django seemed to smooth over that inconsistency when used in filters, but it surfaced quickly once I made the functions model methods.

## Phabricator Ticket
https://phabricator.wikimedia.org/T346103

## How Has This Been Tested?
Local testing + running tests.

## Types of changes
- [X] Minor change (fix a typo, add a translation tag, add section to README, etc.)

### DIFF
--- a/TWLight/applications/views.py
+++ b/TWLight/applications/views.py
@@ -56,7 +56,6 @@ from .helpers import (
     PARTNER_FORM_OPTIONAL_FIELDS,
     PARTNER_FORM_BASE_FIELDS,
     get_output_for_application,
-    count_valid_authorizations,
     get_accounts_available,
     is_proxy_and_application_approved,
     more_applications_than_accounts_available,
@@ -982,11 +981,10 @@ class BatchEditView(CoordinatorsOnly, ToURequired, View):
         # on their valid authorizations to ensure we have enough accounts available and set the
         # boolean value for the corresponding partner_pk in the partners_distribution_flag dictionary.
         for partner_pk, app_count in applications_per_partner.items():
-            total_accounts_available = Partner.objects.filter(pk=partner_pk).values(
-                "accounts_available"
-            )[0]["accounts_available"]
+            this_partner = Partner.objects.get(pk=partner_pk)
+            total_accounts_available = this_partner.accounts_available
             if total_accounts_available is not None:
-                valid_authorizations = count_valid_authorizations(partner_pk)
+                valid_authorizations = this_partner.get_valid_authorization_count
                 total_accounts_available_for_distribution = (
                     total_accounts_available - valid_authorizations
                 )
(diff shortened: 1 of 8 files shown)


Samwalton9:
This is tagged WIP because tests are failing for reasons I can't understand:
```
FAIL: test_create_user_and_editor (TWLight.users.tests.OAuthTestCase)
OAuthBackend._create_user_and_editor() should:
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/usr/local/lib/python3.7/unittest/case.py", line 628, in run
    testMethod()
  File "/usr/local/lib/python3.7/unittest/mock.py", line 1256, in patched
    return func(*args, **keywargs)
  File "/app/TWLight/users/tests.py", line 1654, in test_create_user_and_editor
    self.assertEqual(user.email, "alice@example.com")
  File "/usr/local/lib/python3.7/unittest/case.py", line 852, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/usr/local/lib/python3.7/unittest/case.py", line 1233, in assertMultiLineEqual
    self.fail(self._formatMessage(msg, standardMsg))
  File "/usr/local/lib/python3.7/unittest/case.py", line 693, in fail
    raise self.failureException(msg)
AssertionError: '' != 'alice@example.com'
+ alice@example.com
```

I've dug through the code a bit and can't for the life of me figure out why my changes would result in this test failing.

Samwalton9:
I just checked out master and I'm still getting the same error, which makes me think it's unrelated to this PR and simply a result of the changes I made in the email bug PR.

TWLight/applications/views.py
```diff
@@ -982,11 +981,10 @@ def post(self, request, *args, **kwargs):
         # on their valid authorizations to ensure we have enough accounts available and set the
         # boolean value for the corresponding partner_pk in the partners_distribution_flag dictionary.
         for partner_pk, app_count in applications_per_partner.items():
-            total_accounts_available = Partner.objects.filter(pk=partner_pk).values(
-                "accounts_available"
-            )[0]["accounts_available"]
```
Samwalton9, line 987: I may have been missing something here but this seemed to be an extremely roundabout way of retrieving one value from one object.

jsnshrmn (approved):
This looks good and works in local testing; thanks!

PR #1224, opened 2023-11-18
Add language-adjusted wiki link

suecarmol:
## Description
- Added a property to the UserProfile model to determine which Wikipedia TWL page should be linked in the About section.
- Added a management command to update the languages JSON file to check if any new TWL wiki pages have been added/deleted.

## Rationale
These Wikipedia sites are in the user's preferred language, therefore they might be encouraged to leave comments there

## Phabricator Ticket
[T299510](https://phabricator.wikimedia.org/T299510)

## How Has This Been Tested?
1. Navigate to My Library
2. Check that the {lang_code}.wiki link is in your preferred language
3. Change your preferred language in the settings
4. Check that the {lang_code}.wiki link is in that language
5. If a page does not exist in a user's preferred language, then en.wiki shows up by default

Screenshot 2023-11-17 at 22 13 56
Screenshot 2023-11-17 at 22 16 18

## Types of changes


TWLight/users/models.py
```diff
@@ -148,6 +149,37 @@ def delete_my_library_cache(self):
...
+        wiki_twl_pages = {
+            "en": "https://en.wikipedia.org/wiki/Wikipedia:The_Wikipedia_Library"
+        }
+        page = "Wikipedia:The_Wikipedia_Library"
+        url = "https://en.wikipedia.org/w/rest.php/v1/page/" + page + "/links/language"
+        headers = {
+            "User-Agent": "MediaWiki REST API docs examples/0.1 (https://www.mediawiki.org/wiki/API_talk:REST_API)"
+        }
+
+        response = requests.get(url, headers=headers)
```
jsnshrmn, line 168: this is adding a synchronous backend request to wikipedia as a dependency for every cold load of my library. I suggest using an fixed include list and only returning values that we know are good (eg. active language page or en).

TWLight/users/templates/users/filter_section.html
```diff
@@ -24,6 +24,10 @@
           {% comment %}Translators: On the MyLibrary page (https://wikipedialibrary.wmflabs.org/users/my_library), this appears below introductory text and links to The Wikipedia Library Talk page on meta (https://meta.wikimedia.org/wiki/Talk:The_Wikipedia_Library) {% endcomment %}
           {% trans "Discuss" %}
         </a>
+        &ndash;
+        <a href="{{ user.userprofile.talk_page_i18n }}" class="twl-links twl-links-underline">
+          {{ user.userprofile.lang }}.wiki
```
jsnshrmn, line 29: I find it confusing that the link can display a language-edition wiki when it can take a user to en wiki. I suggest setting it to talk_page_i18n so that the label matches the destination.

jsnshrmn (changes requested):
Thanks for working on this and apologies for the long overdue review! I like that this is really simple, and think you are headed in the right direction. I do have some inline suggestions.

TWLight/users/models.py
suecarmol: I have added a fixed list. Is this what you meant? If so, I would add this list into a JSON file and load its contents. I would also create a small cron job that updates this JSON file with a Wikipedia request a couple of times a year or yearly.

TWLight/users/templates/users/filter_section.html
suecarmol: Done!

TWLight/users/models.py
jsnshrmn: That's good! The code and names are already in the db for user language, so it seems like those should both use the same source of truth for that mapping if feasible (whether they both lookup the db or the json file). But otherwise, yeah, I think just shipping this in a json file like `locale/language-data.json` makes sense!

suecarmol: Done!

TWLight/crons.py
```diff
@@ -89,3 +89,17 @@ def do(self):
             management.call_command("djmail_delete_old_messages", days=100)
         except Exception as e:
             capture_exception(e)
+
+
+class UpdateTwlTalkPage(CronJobBase):
```
jsnshrmn, line 94: I suggest that we keep the management command, but skip the cron task since we're keeping this in git. Otherwise we'll end up with inconsistencies when we deploy a new update that has an older version of the file. We can use a github action to run the update command instead. That could be a followup patch.

suecarmol: Done!

jsnshrmn (approved):

jsnshrmn:
Looks good to me!

PR #1251, opened 2024-01-19
Fix Bristol University Press

amire80:
This was contributed by User:Drkirstyross as a translation into British English (en-gb) on translatewiki, but actually, it's better to make it the default in the default English (en) in the first place.


suecarmol (approved):
Thanks for the correction!

PR #1264, opened 2024-02-08
Homepage collection number update

Samwalton9:
Just a minor update to the number of collections displayed on the homepage and in the embed.


suecarmol (approved):
👍

PR #1276, opened 2024-03-29
Upgrade to Django 4.2

suecarmol:
## Description
Upgrade from Django 3.2 to Django 4.2

## Rationale
This upgrade is necessary because Django 3.2 will no longer have security updates. Django 4.2 is the next LTS version.

## Phabricator Ticket
[T348626](https://phabricator.wikimedia.org/T348626)

## How Has This Been Tested?
Tested manually.

## Types of changes


jsnshrmn (approved):
👍 Thanks for your work! I poked around with this in staging and ran crons locally. Everything I checked worked as expected.

I think that a lot of our routes that were defined using `url()`, which are defined here with `re_path()` simply use `path()` instead. I think `path()` got added somewhere in Django 2.x and we simply never migrated them over since we didn't have to. I think dropping all of the regex that we can is a good thing, but it doesn't have to be in this upgrade. I recommend either creating a follow-up pr on this task or that we file a task for future migration.

PR #1289, opened 2024-05-09
Kgraessle/t357057

katydidnot:

## Description
RTL user experience bugs for The Wikipedia Library

- Fixed twlight_static.sh to generate RTL css by moving the command to run it to the twlight_docker_entrypoint.sh script instead of in the Dockerfile. 
- Added generated files to the .gitignore. 
- Fixed RTL css issues with text getting cut off or overflowing parent container.

Bug: T357057

## Rationale
Fixes The Wikipedia Library RTL bugs reported in https://phabricator.wikimedia.org/T357057. 

## Phabricator Ticket
https://phabricator.wikimedia.org/T357057

## How Has This Been Tested?
Since it was mostly fixing styling issues I did not add tests. 

Yes- start the application with Docker and switch your language settings to a language that reads right to left. 
It's also probably worth manually testing non right to left settings as well. 
Steps to Reproduce: 

1. Start the app and login. 
2. Change your language in the Profile language dropdown to a RTL language (Hebrew and Arabic were the languages I used to test). 
3. Logout. 
4. View home page and ensure the text is not being cut off on the landing page.
5. Log back in to your homepage and resize the window. The two filter headers should not collide in the filter nav. 
6. Go to profile settings page and resize, ensure the two cards with the file icon and directory icon are not overflowing their containers. 
7. Go to the suggestion page, ensure the upvote button is not overflowing it's container. 

### Before
The RTL text was being cutoff like so: 
image

### After
Now, it looks like the following: 
Screenshot 2024-05-10 at 11 25 13 AM

### Before
The second issue I fixed is in the filter panel the 'Filters' and 'Reset Filters' text colliding. 
image (1)

### After 
The text wraps to a new line instead of colliding. 
Screenshot 2024-05-10 at 11 30 04 AM

### Before
The text was overflowing it's parent container 
image (2)

### After 
The text wraps to a new line instead of colliding. 
Screenshot 2024-05-10 at 11 32 21 AM

### Before
The button was overflowing it's parent container. 
image (3)

### After 
Screenshot 2024-05-10 at 11 34 51 AM

## Types of changes


jsnshrmn:
@katydidnot could you add brief manual testing steps and screenshots to the PR description? It doesn't have be anything fancy, just add steps to check each problem and before/after shots for each problem this fixes. See https://github.com/WikipediaLibrary/TWLight/pull/1224 as an example

katydidnot:
> @katydidnot could you add brief manual testing steps and screenshots to the PR description? It doesn't have be anything fancy, just add steps to check each problem and before/after shots for each problem this fixes. See #1224 as an example

Just added photos- hope this helps.

.gitignore
```diff
@@ -179,3 +179,5 @@ tramp
 # cask packages
 .cask/
 
+node_modules/
+TWLight/static/css/*-rtl.css
```
jsnshrmn, line 183: we need the rtl css files in version control for production deployment; git ignoring node_modules makes sense though.

Dockerfile
```diff
@@ -64,8 +69,6 @@ WORKDIR $TWLIGHT_HOME
 
 COPY manage.py /app/manage.py
 
-# Configure static assets.
-RUN SECRET_KEY=twlight /app/bin/twlight_static.sh
```
jsnshrmn, line 68: At some point we had to have this in the image to make our ci process happy; We should go have a look through the github actions for pull requests and production branch merging before making changes.

bin/twlight_docker_entrypoint.sh
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+SECRET_KEY=twlight ./bin/twlight_static.sh
```
jsnshrmn, line 2: I think putting this here means that we have to run the django collect static command any time we restart a container in production. Is that necessary?

jsnshrmn (commented):
This isn't a full review really, but really just focused on the build script part of things.

.gitignore
katydidnot: Interesting, my concern is that if we change new-local.css and then forget to update the corresponding RTL css file then there could be drift between the two.

I kind of like the idea of having a source of truth, that is then "transformed" to generate the RTL css if that makes sense. 
I also assumed that since new-local-rtl.css is not being checked into git and is generated we would want to do the same for the rest of the -rtl.css files, but definitely correct me if I'm wrong. 

I also have little to no context around how deploys to prod go, so take my input with a grain of salt.

bin/twlight_docker_entrypoint.sh
katydidnot: Yes, it would, which I don't like. 

Maybe we can pair up on this on Monday. 
I couldn't get the RTL static assets to generate unless I put it in the entrypoint script after the container has already started.

Dockerfile
katydidnot: Yes, I'd love to look into how the CI/CD process flows!

jsnshrmn:
It looks like the rebase has gone awry; we can look it together tomorrow if you would like.

PR #1290, opened 2024-05-09
any ref with target atribute set to _blank has rel set to noopener an…

katydidnot:
…d noreferrer

## Description

Fix refs with target attribute to include both noopener and noreferrer. 

Bug: T358131

## Rationale
Ensures that links and our website remain secure.

## Phabricator Ticket
https://phabricator.wikimedia.org/T358131

## How Has This Been Tested?
No added tests, but did manually test locally. 
I also fixed tests that broke because of the additional field. 

Yes, by clicking on the search button at the top of The Wikipedia Library. 
You can also find any other link that is opening in a new tab and test that way. 

## Types of changes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -179,3 +179,5 @@ tramp
 # cask packages
 .cask/
 
+node_modules/
+TWLight/static/css/*-rtl.css
--- a/.gitignore
+++ b/.gitignore
@@ -178,6 +178,3 @@ tramp
 
 # cask packages
 .cask/
-
-node_modules/
-TWLight/static/css/*-rtl.css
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -179,3 +179,5 @@ tramp
 # cask packages
 .cask/
 
+node_modules/
+TWLight/static/css/*-rtl.css
```
jsnshrmn, line 183: We need this in version control for production deployments

jsnshrmn (changes requested):
This looks good, but I have one inline comment!

.gitignore
katydidnot: Done!

jsnshrmn (approved):

PR #1291, opened 2024-05-10
Improper resource shutdown or release

katydidnot:
Bug: T358130

## Description

Closing all files that are opened with the open() function when files are done being read. 

## Rationale

> 
Closing files is especially important when accessing file fields in a loop over a large number of objects. If files are not manually closed after accessing them, the risk of running out of file descriptors may arise. This may lead to the following error OSError: [Errno 24] Too many open files

per Django's documentation on [Managing Files](https://docs.djangoproject.com/en/3.2/topics/files/)

## Phabricator Ticket
https://phabricator.wikimedia.org/T358130

## How Has This Been Tested?
I ran all unit tests and ensured they passed as well as performed a smoke test locally. 

## Types of changes

### DIFF
--- a/TWLight/resources/helpers.py
+++ b/TWLight/resources/helpers.py
@@ -186,6 +186,7 @@ def _read_translation_file(language_code: str, filename: str):
     if os.path.isfile(filepath):
         with open(filepath, "r") as translation_file:
             translation_dict = json.load(translation_file)
+
             # Remove the "@metadata" key from the dictionary
             if "@metadata" in translation_dict:
                 translation_dict.pop("@metadata")
--- a/TWLight/resources/helpers.py
+++ b/TWLight/resources/helpers.py
@@ -186,7 +186,7 @@ def _read_translation_file(language_code: str, filename: str):
     if os.path.isfile(filepath):
         with open(filepath, "r") as translation_file:
             translation_dict = json.load(translation_file)
-
+            translation_file.close()
             # Remove the "@metadata" key from the dictionary
             if "@metadata" in translation_dict:
                 translation_dict.pop("@metadata")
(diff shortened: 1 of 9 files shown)


TWLight/resources/helpers.py
```diff
@@ -186,7 +186,7 @@ def _read_translation_file(language_code: str, filename: str):
     if os.path.isfile(filepath):
         with open(filepath, "r") as translation_file:
             translation_dict = json.load(translation_file)
-
+            translation_file.close()
```
jsnshrmn, line 189: I think the files that we open in `with` blocks are closed automatically

syslog/bin/import_logs.py
```diff
@@ -938,7 +938,7 @@ def _parse_args(self, option_parser, argv = None):
 
         if self.options.output:
             sys.stdout = sys.stderr = open(self.options.output, 'a+', 0)
-
+            self.options.output.close()
```
jsnshrmn, line 941: I think this needs to stay open for the lifetime of the script execution for it to append logs as expected.

jsnshrmn (commented):
Thanks for your work! comments inline

TWLight/resources/helpers.py
katydidnot: You are correct. I totally read the Django docs incorrectly. **Always read the second sentence**. 🤦‍♀️ 
Screenshot 2024-05-10 at 11 44 36 AM

syslog/bin/import_logs.py
katydidnot: I moved the file closures to after main() method runs and before we exit 0.

syslog/bin/import_logs.py
```diff
@@ -1115,6 +1113,7 @@ def _get_token_auth(self):
 
             filename = stdout
             credentials = open(filename, 'r').readline()
+            filename.close()
```
katydidnot, line 1116: Should this be closed here?

jsnshrmn: Yep, though the risk is super low since this code isn't web-accessible

jsnshrmn (approved):
This looks good and doesn't impact functionality in any way. Thanks for the patch!

PR #1293, opened 2024-05-11
Show block criteria success if user is exempted

Samwalton9:

## Description
Some users get confused when they have received a block exemption but then fall below the activity thresholds for access to the library. In doing so they still see the message about failing the block criterion, and email us asking why their exemption is no longer active. In an ideal world we could communicate something about the exemption here, but in the meantime it's less confusing to just show that they succeeded against this criterion instead of failed.

This change both shows a success and hides the extra message about contacting us.

## Phabricator Ticket
[T327489](https://phabricator.wikimedia.org/T327489)

## How Has This Been Tested?
Ran tests locally, manually tested with a blocked & exempted second account.

## Types of changes

### DIFF
--- a/TWLight/users/templates/users/eligibility_modal.html
+++ b/TWLight/users/templates/users/eligibility_modal.html
@@ -33,7 +33,7 @@
           </p>
         {% endif %}
       </div>
-      {% if not editor.wp_not_blocked %}
+      {% if not editor.wp_not_blocked and not user.editor.ignore_wp_blocks %}
         {% url 'contact' as contact_url %}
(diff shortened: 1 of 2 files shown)


katydidnot (approved):
This looks good to me. 

All tests are passing. I also tested manually with editing my user to be blocked so that I could see the block message, then I set ignore_wp_blocks to true and verified that I no longer saw the block message.

suecarmol (approved):
Tested this and it does what it says in the ticket! Thanks, volunteer Sam!

PR #1297, opened 2024-05-17
Kgraessle/t226804

katydidnot:

## Description

Allow users to change their interface language from any page

Bug: T226804

## Rationale

- A user that is not logged in can change their language and have it persist even after they login. 
- Whenever a user is navigating the application they are able to change the language setting from any screen.  

## Phabricator Ticket
[T226804](https://phabricator.wikimedia.org/T226804)

## How Has This Been Tested?

I added unit tests to cover the language logic in the method get_and_update_user_from_identity().
I also backfilled some unit tests for the method check_user_preferred_language() since I was building upon it. 

## Not Logged In: 
Screenshot 2024-05-17 at 12 26 17 PM
Screenshot 2024-05-17 at 12 26 23 PM

## Not Logged In RTL: 
Screenshot 2024-05-17 at 12 26 00 PM

## Logged in: 
Screenshot 2024-05-17 at 12 25 36 PM
Screenshot 2024-05-17 at 12 25 28 PM

## Logged in RTL: 
Screenshot 2024-05-17 at 12 25 50 PM

## Types of changes
- [X] New feature (non-breaking change which adds functionality)


suecarmol (changes requested):
Thank you for working on this! I have tested the code and it looks good, but there is something that needs to be fixed before we can merge this.

Some dropdown language options are translated when you change the language. This should not happen. The behavior should be the same as the dropdown in `/users`.

Screenshot 2024-05-20 at 18 50 54

Screenshot 2024-05-20 at 18 51 30

TWLight/templates/header_partial_b4.html
```diff
@@ -61,9 +61,31 @@
...
+          <li class="mobile-menu align-middle nav-item">
+            <form action="/i18n/setlang/?next=/" method="post" class="form-inline">
+                {% csrf_token %}
+                {% if form_show_errors %}
+                  {% include "bootstrap4/errors.html" %}
+                {% endif %}
+                <div class="dropdown">
+                  <button class="btn btn-link dropdown-toggle nav-link nav-links language-select-button" type="button" id="language-dropdown" data-toggle="dropdown"
+                          aria-haspopup="true" aria-expanded="false">
+                          {% for code, name in LANGUAGES %}
```
katydidnot, line 76: I'd like to DRY this up a bit by not looping twice. 
The only problem is we've added additional locales not supported by Django's built-in helper function get_language_info so I'm not sure how else to grab localized names without looping through all the options and checking for equality. 

Any ideas or thoughts?

suecarmol: You are right, we can do this without looping. Try https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#get-language-info.

katydidnot: Yeah, I had tried that method and was getting an exception locally for locales that aren't supported out of the box by Django. 

For example, Bikol Central returns the following error: 
Screenshot 2024-05-22 at 6 13 45 AM

suecarmol: Based on that, I think we can leave this as-is. I hope Django will one day catch-up on what languages they support.

suecarmol (approved):
This looks good and works in different screen sizes. Great job!

PR #1303, opened 2024-05-30
Remove unnecessary whitespace in tooltip

Justman100:


locale/en/LC_MESSAGES/django.po
```diff
@@ -2346,7 +2346,7 @@ msgstr ""
 
 #. Translators: Alt text for the Wikipedia Library shown in the top left of all pages.
 #: TWLight/templates/login_partial.html:12
-msgid " The Wikipedia Library"
+msgid "The Wikipedia Library"
```
suecarmol, line 2349: You don't have to modify this file. Our CI will update all django.po messages

suecarmol (changes requested):
Thank you for spotting this!

PR #1307, opened 2024-06-17
Fix translation string error in /terms

suecarmol:
## Description
Fixes a translation string in Spanish which caused the `/terms` page to error out.

## Rationale
This fixes the `/terms` page, which new users who picked Spanish as their UI language must agree to before using the library.

## Phabricator Ticket
[T366620](https://phabricator.wikimedia.org/T366620)

## How Has This Been Tested?
Tested manually that the page is loading correctly in Spanish and in other languages.

## Types of changes

### DIFF
--- a/TWLight/users/oauth.py
+++ b/TWLight/users/oauth.py
@@ -263,8 +263,9 @@ class OAuthBackend(object):
         # The authenticate() function of a Django auth backend must return
         # the user.
         return user
-    
-    # Implementation for https://docs.djangoproject.com/en/4.2/ref/contrib/auth/#django.contrib.auth.get_user
+
+    # Implementation for
+    # https://docs.djangoproject.com/en/4.2/ref/contrib/auth/#django.contrib.auth.get_user
     def get_user(self, user_id):
         user = User.objects.filter(pk=user_id).first()
         if user is None:
--- a/TWLight/users/oauth.py
+++ b/TWLight/users/oauth.py
@@ -180,6 +180,7 @@ class OAuthBackend(object):
             if should_update_lang:
                 user.userprofile.lang = get_language()
                 user.userprofile.save()
+
             # This login path should only be used for accounts created via
             # Wikipedia login, which all have editor objects.
             if hasattr(user, "editor"):
@@ -264,10 +265,11 @@ class OAuthBackend(object):
         return user
 
     def get_user(self, user_id):
-        try:
-            return User.objects.get(pk=user_id)
-        except User.DoesNotExist as e:
-            logger.exception(e)
+        user = User.objects.filter(pk=user_id).first()
+        if user is not None:
+            return user
+        else:
+            logger.exception("OAuthBackend.get_user: User does not exist")
             return None
 
 
--- a/TWLight/users/oauth.py
+++ b/TWLight/users/oauth.py
@@ -266,11 +266,9 @@ class OAuthBackend(object):
 
     def get_user(self, user_id):
         user = User.objects.filter(pk=user_id).first()
-        if user is not None:
-            return user
-        else:
-            logger.exception("OAuthBackend.get_user: User does not exist")
-            return None
+        if user is None:
+            logger.warning("OAuthBackend.get_user: User does not exist")
+        return user
 
 
 class OAuthInitializeView(View):
--- a/TWLight/users/oauth.py
+++ b/TWLight/users/oauth.py
@@ -263,7 +263,8 @@ class OAuthBackend(object):
         # The authenticate() function of a Django auth backend must return
         # the user.
         return user
-
+    
+    # Implementation for https://docs.djangoproject.com/en/4.2/ref/contrib/auth/#django.contrib.auth.get_user
     def get_user(self, user_id):
         user = User.objects.filter(pk=user_id).first()
         if user is None:


TWLight/users/oauth.py
```diff
@@ -264,10 +265,11 @@ def authenticate(self, request=None, access_token=None, handshaker=None):
...
     def get_user(self, user_id):
-        try:
-            return User.objects.get(pk=user_id)
-        except User.DoesNotExist as e:
-            logger.exception(e)
+        user = User.objects.filter(pk=user_id).first()
+        if user is not None:
+            return user
+        else:
+            logger.warning("OAuthBackend.get_user: User does not exist")
```
jsnshrmn, line 272: I had to go lookup the django docs for how this method is supposed to work.
https://docs.djangoproject.com/en/4.2/ref/contrib/auth/#django.contrib.auth.get_user

I'm pretty sure I created the no return problem here years ago. We should probably document in this class that it's implementing `django.contrib.auth`, just as a reminder to ourselves.
suggestion for legibility
```suggestion
        user = User.objects.filter(pk=user_id).first()
        if user is None:
            logger.warning("OAuthBackend.get_user: User does not exist")
        return user         
```

jsnshrmn (commented):
Thanks for your work on this! Can you explain the translation changes? Typically, if we have an issue with a message string, we change the message in source and then let cicd handle updates to the i18n files in locale.

TWLight/users/oauth.py
```diff
@@ -264,10 +265,11 @@ def authenticate(self, request=None, access_token=None, handshaker=None):
...
             return None
```
jsnshrmn, line 273: I had to go lookup the django docs for how this method is supposed to work.
https://docs.djangoproject.com/en/4.2/ref/contrib/auth/#django.contrib.auth.get_user

I'm pretty sure I created the no return problem here years ago. We should probably document in this class that it's implementing `django.contrib.auth`, just as a reminder to ourselves.
suggestion for legibility
```suggestion
        user = User.objects.filter(pk=user_id).first()
        if user is None:
            logger.warning("OAuthBackend.get_user: User does not exist")
        return user         
```
```suggestion
        user = User.objects.filter(pk=user_id).first()
        if user is not None:
            return user
        else:
            logger.warning("OAuthBackend.get_user: User does not exist")
            return None
```

suecarmol: Gotcha! I changed the logger from `exception` to `warning` because it was throwing errors when it shouldn't have and it is making noise in our GlitchTip.

locale/en/LC_MESSAGES/django.po
```diff
@@ -2517,82 +2517,82 @@ msgid ""
 msgstr ""
 
 #. Translators: Do not translate.
-#: TWLight/users/models.py:793
+#: TWLight/users/models.py:794
```
katydidnot, line 2520: What does the number incrementing mean? Is this just a reference to a line of code?

jsnshrmn: Yep, there is an entry for every place the string appears. These comments are generated automatically.

katydidnot (changes requested):
I had the same question Jason had about the messages being populated from CI/CD. 

And a minor fix:  
When I ran the tests there was a linter error. 
Here's the output:

``` docker compose exec --no-TTY twlight /app/bin/virtualenv_test.sh
WARN[0000] /Users/katherinegraessle/Workspace/TWLight/docker-compose.yml: `version` is obsolete
WARN[0000] /Users/katherinegraessle/Workspace/TWLight/docker-compose.override.yml: `version` is obsolete
[Tue Jun 18 15:26:31 UTC 2024]
black --target-version py311 --check TWLight
would reformat /app/TWLight/users/oauth.py

Oh no! 💥 💔 💥
1 file would be reformatted, 245 files would be left unchanged.
--- /app/TWLight/users/oauth.py	2024-06-18 15:00:44.018872+00:00
+++ /app/TWLight/users/oauth.py	2024-06-18 15:26:33.123470+00:00
@@ -261,17 +261,17 @@
         request.session["user_created"] = created

         # The authenticate() function of a Django auth backend must return
         # the user.
         return user
-
+
     # Implementation for https://docs.djangoproject.com/en/4.2/ref/contrib/auth/#django.contrib.auth.get_user
     def get_user(self, user_id):
         user = User.objects.filter(pk=user_id).first()
         if user is None:
             logger.warning("OAuthBackend.get_user: User does not exist")
-        return user
+        return user


 class OAuthInitializeView(View):
     """
     Ask Wikipedia for a temporary key/secret for the user, and redirect
You can fix these issues by running the following command on your host
docker exec CONTAINER /venv/bin/black -t py311 /app/TWLight ```

**Comment by suecarmol** [comment]
> Thanks for your work on this! Can you explain the translation changes? Typically, if we have an issue with a message string, we change the message in source and then let cicd handle updates to the i18n files in locale.

I think the translation changes happened when I ran `bin/virtualenv_translate.sh force`

**Comment by jsnshrmn** [comment]
> > Thanks for your work on this! Can you explain the translation changes? Typically, if we have an issue with a message string, we change the message in source and then let cicd handle updates to the i18n files in locale.
> 
> I think the translation changes happened when I ran `bin/virtualenv_translate.sh force`

That makes sense; can you drop them from this pr?

**Comment by suecarmol** [comment]
> > > Thanks for your work on this! Can you explain the translation changes? Typically, if we have an issue with a message string, we change the message in source and then let cicd handle updates to the i18n files in locale.
> > 
> > 
> > I think the translation changes happened when I ran `bin/virtualenv_translate.sh force`
> 
> That makes sense; can you drop them from this pr?

Ok, after about an hour of fighting Git, I think this is ready

**Review by katydidnot — APPROVED** [review]
I think this looks good, but just had one question to help me butter understand the process: does the updated translation from the messages file auto-populate in translatewiki when you change it?

**Comment by jsnshrmn** [comment]
> I think this looks good, but just had one question to help me butter understand the process: does the updated translation from the messages file auto-populate in translatewiki when you change it?

We update the strings in our application code. Then our CI runs a job that extracts all of the messages for i18n. Then translatewiki pulls those messages. Then people update translatewiki. Then translate wiki commits new translations to our message files. But we should never update the non english non-qqq messages in master. And those messages (english and qqq) we should only update indirectly by updating our strings and comments in our source code.

**Comment by katydidnot** [comment]
> Then translate wiki commits new translations to ou

Thank you, this is helpful.

File: locale/es/LC_MESSAGES/django.po [hunk 4]
```diff
@@ -2841,7 +2841,7 @@ msgstr "Algunas páginas de La Biblioteca de Wikipedia son navegables in tener q
...
+msgstr "Para administrar el programa de la Biblioteca de Wikipedia, conservaremos los datos de la aplicación que recopilamos de usted durante tres años después de su inicio de sesión más reciente, a menos que elimine su cuenta, como se describe a continuación. Puede iniciar sesión e ir a <a class=\"twl-links\" href=\"%(profile_url)s\">su página de perfil</a> para ver la información asociada con su cuenta y descargarla en formato JSON. Puede acceder, actualizar, restringir o eliminar esta información en cualquier momento, excepto la información que se recupera automáticamente de los proyectos. Si tiene alguna pregunta o inquietud sobre el manejo de sus datos, comuníquese con <a class=\"twl-links\" href=\"mailto:wikipedialibrary@wikimedia.org?Subject=Wikipedia%%20Library%%20Card%%20account%%20deactivation\"> wikipedialibrary@wikimedia.org</a>."
```
jsnshrmn, line 2844: So this was definitely our problem, but I'm not sure that we should be updating this directly.

jsnshrmn: We should be fixing things like this in translatewiki. If this string doesn't match what's in translatewiki, then there is something weird going on.

suecarmol: It worked on staging. I've already fixed the translation in TranslateWiki, so we can wait for Thursday's update.

PR #1320, opened 2024-07-12
OAuth: User.DoesNotExist is not being caught in try/except

katydidnot:
Bug: T368663

## Description

Better error handling for users that do not exist. 

## Rationale

- Glitchtip errors
- In the oauth.py flow, a try/except clause searches for the user with User.objects.get() that should catch when a user doesn't exist. That doesn't seem to happen, so we should change how we get the user. We should change the logic to User.objects.filter().first() and then check if the user is None.

## Phabricator Ticket
https://phabricator.wikimedia.org/T368663

## How Has This Been Tested?
Ran unit tests to ensure no existing functionality broke.
Tested login flow locally. 

##  Can this change be tested manually? How?
Yes, ensure that you can log in as a new user and an existing user.

## Types of changes

### DIFF
--- a/TWLight/resources/templates/resources/suggest.html
+++ b/TWLight/resources/templates/resources/suggest.html
@@ -157,7 +157,12 @@
       }
       var this_ = $(this);
       var upvoteURL = this_.attr("data-href");
-      $.get(upvoteURL);
+      $.ajax({
+        url: upvoteURL,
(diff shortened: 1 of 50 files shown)


suecarmol (changes requested):
Please rebase this PR so that the translate wiki changes are not included.

PR #1343, opened 2024-08-20
Delete previous sessions when a user logs in

suecarmol:
## Description
- Delete previous sessions when a user logs in
- Added filters and search in sessions admin interface

## Rationale
Users may have multiple sessions open and can access pages simultaneously without
the user being notified, which extends the time available for the attacker to perform illegitimate
actions on the platform on their behalf.

## Phabricator Ticket
[T358124](https://phabricator.wikimedia.org/T358124)

## How Has This Been Tested?
1. Log-in in a browser tab. Do not log out
2. Open a private/different session in your browser and log in
3. Refresh the first browser tab. You should be logged out
4. Check in /admin that there is only one session per user

## Types of changes

### DIFF
--- a/TWLight/users/oauth.py
+++ b/TWLight/users/oauth.py
@@ -24,7 +24,7 @@ from django.utils.translation import (
 
 from urllib.parse import urlencode
 
-from .models import Editor
+from .models import Editor, Session
 
 
 logger = logging.getLogger(__name__)
@@ -271,6 +271,12 @@ class OAuthBackend(object):
             logger.info("User has been updated.")
 
         request.session["user_created"] = created
+
+        # Checking if more than one session exists
+        sessions = Session.objects.filter(account_id=user.pk)
+        if sessions.count() >= 1:
+            logger.info("Deleting all previous sessions.")
+            sessions.delete()
 
         # The authenticate() function of a Django auth backend must return
         # the user.
@@ -524,7 +530,6 @@ class OAuthCallbackView(View):
                     # fmt: on
                 ),
             )
-            local_redirect = reverse_lazy("homepage")
 
         user = authenticate(
             request=request, access_token=access_token, handshaker=handshaker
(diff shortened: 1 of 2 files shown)


TWLight/users/oauth.py
```diff
@@ -272,6 +272,12 @@ def authenticate(self, request=None, access_token=None, handshaker=None):
 
         request.session["user_created"] = created
 
+        # Checking if more than one session exists
+        sessions = Session.objects.filter(account_id=user.pk)
+        if sessions.count() >= 1:
+            logger.info("Deleting all previous sessions.")
+            sessions.delete()
```
katydidnot, line 279: I have a question about deleting all sessions by account_id-  will this affect the user's current session?

katydidnot (commented):
This looks pretty good to me, I just had one question about deleting all sessions.

TWLight/users/oauth.py
suecarmol: It does not, otherwise, you would enter an infinite log-in loop. I also thought that at first, so I excluded the latest session from my delete query, but two sessions would remain in the /admin. I think the session is created a bit after authentication.

katydidnot: Ok, sounds good, that's what I was thinking too but wanted to verify!

katydidnot:
Approved, but will let Jason take a look first before merging.

TWLight/users/oauth.py
```diff
@@ -272,6 +272,12 @@ def authenticate(self, request=None, access_token=None, handshaker=None):
 
         request.session["user_created"] = created
 
+        # Checking if more than one session exists
+        sessions = Session.objects.filter(account_id=user.pk)
+        if sessions.count() >= 1:
```
jsnshrmn, line 277: question: could this be written as?
```suggestion
        if sessions.exists():
```

jsnshrmn (approved):
This looks good to me, though I do have one inline question (the only reason I'm not merging immediately). It's not a blocker, so feel free to resolve it and merge if you don't believe any change is warranted. Thanks for your work!

TWLight/users/oauth.py
suecarmol: That's true! This was a leftover from when I thought I needed to deleted all sessions but the latest.

PR #1402, opened 2025-01-10
Fix broken HTML on East View Press page

Samwalton9:
We accidentally added an extra opening `<strong>` tag which is causing the East View Press page to break
Screenshot 2025-01-10 at 12 54 44

Removing all the strong tags that get added as a result in Inspect fixes the page:
Screenshot 2025-01-10 at 12 55 33


jsnshrmn (approved):
👍

PR #1418, opened 2025-02-11
Update Emerald description

Samwalton9:
Emerald are only providing access to one collection now, so update the description to reflect that.


suecarmol (approved):
👍

PR #1453, opened 2025-06-28
Add management command that sends monthly reports of logged-in users

suecarmol:
## Description
Added a management command that retrieves user names that have logged in in the past month and have approved applications and current authorizations.

## Rationale
Currently, we cannot provide monthly breakdowns of users who have logged into TWL and accessed its content.

## Phabricator Ticket
[T397222](https://phabricator.wikimedia.org/T397222)

## How Has This Been Tested?
Tested manually by running `python manage.py retrieve_monthly_users`

## Types of changes

### DIFF
--- a/TWLight/emails/templates/emails/user_retrieve_monthly_logins-body-html.html
+++ b/TWLight/emails/templates/emails/user_retrieve_monthly_logins-body-html.html
@@ -1,0 +1,9 @@
+<html>
+<body>
+<p> Hi Vipin, </p>
+<p> Here is a list of the users that logged-in last month that have approved
+  applications and current authorizations. </p>
+
+<p> {{ monthly_users|json_script:"monthly-users" }} </p>
+</body>
+</html>
--- a/TWLight/emails/templates/emails/user_retrieve_monthly_logins-body-text.html
+++ b/TWLight/emails/templates/emails/user_retrieve_monthly_logins-body-text.html
@@ -1,0 +1,3 @@
+Hi Vipin,
+Here is a list of the users that logged-in last month that have approved applications and current authorizations.
+{{ monthly_users|json_script:"monthly-users" }}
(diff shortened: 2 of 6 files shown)


TWLight/emails/templates/emails/user_retrieve_monthly_logins-body-html.html
```diff
@@ -0,0 +1,9 @@
+<html>
+<body>
+<p> Hi Vipin, </p>
```
katydidnot, line 3: Nit: maybe remove Vipin's name in case we end up sending it to someone else too? 

Question: 

I tested this locally and I didn't see any errors but I also didn't see anything happen in the console. 
Is this expected behavior? Also where do we configure the email address we're sending it to, is this on the server? 

Otherwise, this looks good.

TWLight/emails/templates/emails/user_retrieve_monthly_logins-body-text.html
```diff
@@ -0,0 +1,3 @@
+Hi Vipin,
```
katydidnot, line 1: nit: same, maybe remove Vipin's name in case we end up sending it to other people as well.

TWLight/emails/templates/emails/user_retrieve_monthly_logins-body-html.html
suecarmol: In your local environment, you should be able to see the body of the email printed in the logs.

suecarmol: > Also where do we configure the email address we're sending it to, is this on the server?

I used an environment variable of an existing email, but we could create a new one in .env and change it to the value we want in production.

katydidnot: Hmm, I'm not seeing anything output to the logs. 
Am I missing any other local setup?
I verified that I have a record in my `users_authorization` table .

PR #1465, opened 2025-08-15
POC: Add email sender with API:Emailuser

suecarmol:
## Description
- Removed the `version` item in all docker-compose files (the warning was bothering me)
- Created the send_email_via_api command for other services to send emails from

Caveats from this approach:
- Rate-limiting handling had not been implemented
- If TWL users have a different email address from their Wikipedia account, it could be unwelcome to send an email to that address
- If a TWL user does not have the `Allow other users to email me` option set in Special:Preferences, then they will not receive an email

## Rationale
Our email sending system has issues when sending to certain accounts. This is a possible solution to that problem.

## Phabricator Ticket
[T383070](https://phabricator.wikimedia.org/T383070)

## How Has This Been Tested?
Tested locally. You can run `python manage.py retrieve_monthly_users` and change the `target` parameter to your Wikipedia username.

## Types of changes

### DIFF
--- a/TWLight/emails/management/commands/send_email_via_api.py
+++ b/TWLight/emails/management/commands/send_email_via_api.py
@@ -1,0 +1,125 @@
+import logging
+import os
+import requests
+
+from django.core.management.base import BaseCommand, CommandError
+
+logger = logging.getLogger("django")
+
+
+class Command(BaseCommand):
+    help = "Command that sends emails via API:Emailuser in MediaWiki"
+
+    def info(self, msg):
+        # Log and print so that messages are visible
+        # in docker logs (log) and cron job logs (print)
+        logger.info(msg)
+        self.stdout.write(msg)
+        self.stdout.flush()
+
+    def add_arguments(self, parser):
+        # Named (optional) arguments
+        parser.add_argument(
+            "--target",
+            type=str,
+            help="The Wikipedia username you want to send the email to.",
+        )
+        parser.add_argument(
+            "--subject",
+            type=str,
+            help="The subject of the email.",
+        )
+        parser.add_argument(
+            "--body",
+            type=str,
+            help="The body of the email.",
+        )
+
+    def handle(self, *args, **options):
+        if not options["target"]:
+            raise CommandError("You need to specify a user to send the email to")
+
+        if not options["subject"]:
+            raise CommandError("You need to specify the subject of the email")
+
+        if not options["body"]:
+            raise CommandError("You need to specify the body of the email")
+
+        target = options["target"]
+        subject = options["subject"]
+        body = options["body"]
+
+        email_bot_username = os.environ.get("EMAILWIKIBOTUSERNAME", None)
+        email_bot_password = os.environ.get("EMAILWIKIBOTPASSWORD", None)
+
+        if email_bot_username is None or email_bot_password is None:
+            # Bot credentials not provided; exiting gracefully
+            raise CommandError(
+                "The email bot username and/or password were not provided. Exiting..."
+            )
+        # Code taken in part from https://www.mediawiki.org/wiki/API:Emailuser#Python
+        session = requests.Session()
+        # TODO: See if we need to change this to Meta or the user's home wiki?
+        # Or is this wiki fine?
+        url = "https://test.wikipedia.org/w/api.php"
+
+        # Step 1: GET request to fetch login token
+        login_token_params = {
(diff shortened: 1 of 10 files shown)


TWLight/emails/management/commands/send_email_via_api.py
```diff
@@ -0,0 +1,125 @@
...
+
+        self.info("Signing in...")
+        login_response = session.post(url, data=login_params)
+        if login_response.status_code != 200:
+            raise CommandError("There was an error in the request for the login.")
+
+        # Step 3: GET request to fetch Email token
+        email_token_params = {"action": "query", "meta": "tokens", "format": "json"}
+
+        self.info("Getting emwail token...")
```
katydidnot, line 106: nit: typo in email here

TWLight/emails/management/commands/send_email_via_api.py
```diff
@@ -0,0 +1,125 @@
...
+            "action": "emailuser",
+            "target": target,
+            "subject": subject,
+            "text": body,
+            "token": email_token,
+            "format": "json",
+        }
+
+        self.info("Sending email...")
+        email_response = session.post(url, data=email_params)
```
katydidnot, line 123: I'm getting a 200 back from this request, but I'm also not seeing any emails being sent to my inbox. Is there any setup I need to do to get this to work? I made sure my preferences allowed other users to email me.

TWLight/emails/management/commands/send_email_via_api.py
suecarmol: Good catch!

TWLight/users/management/commands/retrieve_monthly_users.py
```diff
@@ -44,12 +54,17 @@ def handle(self, *args, **options):
...
 
         if monthly_users:
-            UserLoginRetrieval.user_retrieve_monthly_logins.send(
-                sender=self.__class__, monthly_users=monthly_users
+            self.info("Sending email...")
+            call_command(
+                "send_email_via_api",
+                target="Suecarmol",  # TODO: change to Vipin's user
+                subject="Monthly user report",
+                body=f"Here is a list of monthly users: {monthly_users}",
```
katydidnot, line 69: nit: I would put a new line character between the list and the "Here is a list of monthly users:" string.

katydidnot (approved):
Thanks! This looks like a great POC and works as expected, just had some minor nits.

TWLight/users/management/commands/retrieve_monthly_users.py
suecarmol: Done!

jsnshrmn:
Thanks for working on this!

>Caveats from this approach:

I don't think these are intrinsic to the approach, just this implementation;

>    Rate-limiting handling has not been implemented

The need for that will be determined by the account that owns the bot; 
It looks like autoconfirmed users aren't rate limited? There's also a specific user right for skipping rate limits: `noratelimit`
see https://www.mediawiki.org/wiki/Manual:$wgRateLimits
and
https://gerrit.wikimedia.org/r/plugins/gitiles/operations/mediawiki-config/+/refs/heads/master/wmf-config/InitialiseSettings.php#3214

>    If TWL users have a different email address from their Wikipedia account, it could be unwelcome to send an email to that address
>    If a TWL user does not have the Allow other users to email me option set in Special:Preferences, then they will not receive an email

I think we can add error handling for these things:

https://www.mediawiki.org/wiki/API:Emailuser#Possible_errors

I would suggest that we implement a warning email in TWLight for users that meet these cases; letting them know that if they want to continue to receive email from us (and participate in partners that require email), they need to change their wiki settings.

We could either just prepend a warning banner template to every email for users that meet these conditions or add account flags to track the issue at login and have a separate command for sending a standalone email to impacted users.

I think this mostly proves out the TWL engineering side of things; I think to close out this task, we get feedback on how to handle the above situations and figure out how we would want / need to do the on-wiki account setup for the bot owner.

TWLight/emails/management/commands/send_email_via_api.py
jsnshrmn: This isn't too critical for the poc since it's working for the account y'all have setup, but I haven't actually gotten this to work despite setting up bots with "email other users" enabled. Note that in both cases, either the source or the target had special characters in the username, so it could simply be a text escaping issue.

PR #1474, opened 2025-09-12
Add dir attribute to collection tile descriptions

DHardy-WMF:
## Description
Added `dir="ltr"` and `dir="rtl"` attributes to the `p.card-text` element.

## Rationale
Fixes issues where the user's language is RTL and the directionality of partner descriptions may be mixed.

## Phabricator Ticket
https://phabricator.wikimedia.org/T277139

## How Has This Been Tested?
Local

Screenshot 2025-09-12 at 12 01 10 PM
Screenshot 2025-09-12 at 12 01 55 PM

## Types of changes

### DIFF
--- a/TWLight/resources/helpers.py
+++ b/TWLight/resources/helpers.py
@@ -5,22 +5,32 @@ import logging
 import os
 
 
-"""
-Caches the direction (LTR/RTL) of each language by checking if the
-first list value (script type) is defined in the RTL list (rtlscripts)
-"""
-try:
-    with open(os.path.join(settings.LOCALE_PATHS[0], "language-data.json")) as file:
-        data = json.load(file)
-        languages = data["languages"]
-        rtlscripts = data["rtlscripts"]
-        LANGS_DIRECTION = {
-            lang: "rtl" if meta[0] in rtlscripts else "ltr"
-            for lang, meta in languages.items()
-        }
-except Exception as e:
-    logging.getLogger(__name__).exception("Failed to load language data")
-    LANGS_DIRECTION = {}
+LANGS_DIRECTION = {}
+
+
+def get_language_direction(language_code: str):
+    """
+    Caches the direction (LTR/RTL) of each language by checking if the
+    first list value (script type) is defined in the RTL list (rtlscripts)
+    """
+    if not LANGS_DIRECTION:
+        try:
+            with open(
+                os.path.join(settings.LOCALE_PATHS[0], "language-data.json")
+            ) as file:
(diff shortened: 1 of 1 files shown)


katydidnot (changes requested):
Thanks for your work on this, it's working as expected locally. 

I had a few comments and I also was wondering if this is something we can add some unit tests for?

jsnshrmn (approved):
This looks like a nice improvement to me; thanks for putting a fresh set of eyes on it!

suecarmol (commented):
This might be out of scope for this patch, but it would also be useful if we could change all instances of `LANGUAGES_BIDI` to include the RTL languages that are unsupported by Django. This is a GH code search for that term: https://github.com/search?q=repo%3AWikipediaLibrary%2FTWLight%20BIDI&type=code

DHardy-WMF:
> This might be out of scope for this patch, but it would also be useful if we could change all instances of `LANGUAGES_BIDI` to include the RTL languages that are unsupported by Django. This is a GH code search for that term: https://github.com/search?q=repo%3AWikipediaLibrary%2FTWLight%20BIDI&type=code

I agree, we should have parity elsewhere as the `languages-data.json` file has far more detail than what Django provides. But I think it should be covered in another task, since further changes would end up affecting all pages.

TWLight/resources/helpers.py
```diff
@@ -1,9 +1,28 @@
 from django.conf import settings
 from numbers import Number
 import json
+import logging
 import os
 
 
+"""
```
suecarmol, line 8: Let's put this into a function. This helper file has several functions that are unrelated to translation code. Whenever we import this file, this segment will run regardless of whether it is needed or not. If we move this to a function, we can call it as needed.

suecarmol (commented):
I have one more nit. Otherwise, this looks good after testing in staging.

suecarmol (approved):
Thank you for making these changes.

DHardy-WMF:
> > This might be out of scope for this patch, but it would also be useful if we could change all instances of `LANGUAGES_BIDI` to include the RTL languages that are unsupported by Django. This is a GH code search for that term: https://github.com/search?q=repo%3AWikipediaLibrary%2FTWLight%20BIDI&type=code
> 
> I agree! We should have parity elsewhere as the `languages-data.json` file has far more detail than what Django provides. But I think it should be covered in another task, since further changes would end up affecting all pages.

Follow up task created: https://phabricator.wikimedia.org/T404882

PR #1535, opened 2026-06-29
Tests: Use .copy() to prevent mutation of shared module-level test data

theresnotime:
My thinking so far is when tests run in a certain order (especially with `--parallel`), the email-related tests would mutate `FAKE_IDENTITY["email"]` to `None` or `""` _before_ `test_create_user_and_editor` ran, causing its `assertEqual(user.email, "alice@example.com")` assertion to fail.

Bug: [T352896](https://phabricator.wikimedia.org/T352896)

### DIFF
--- a/TWLight/users/tests.py
+++ b/TWLight/users/tests.py
@@ -1692,8 +1692,8 @@ class EditorModelTestCase(TestCase):
         platform, we shouldn't be overwriting it with a blank string.
         """
         new_editor = EditorFactory(wp_registered=None)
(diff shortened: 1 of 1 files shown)


suecarmol (approved):
Thank you for taking the time to fix this bug!

suecarmol:
Merged and getting the flaky error: `ValueError: Model instances passed to related filters must be saved.` See https://github.com/WikipediaLibrary/TWLight/actions/runs/28561755715/job/84680651811 for more context

PR #1537, opened 2026-07-01
Suggested wording for a few labels

framawiki:
Hello!

I'm working these days to present TWL to various new contributors.

Every time I go on the portal I notice that some of the small text labels could be improved. So I'd like to offer a few suggestions for improvement. Feel free to rather use any other ideas you may have!

By the way, I still don't understand what "Last reviewed on Aug. 31, 2025, 7:49 p.m." means on the "My applications" page, even after looking at the code. Maybe someone who knows more about this can suggest a better explanation of this label?

Bug: [T430924](https://phabricator.wikimedia.org/T430924)


suecarmol (changes requested):
Thank you for working to improve the library! Any changes on the library should be linked to a Phabricator task. If you don't see a task relevant to the changes you want to make, you can create a ticket at https://phabricator.wikimedia.org/tag/library-card-platform/.  While some of the changes in the PR might be good changes, others are wordier and could make the UI harder to understand. I encourage you to create a Phabricator task where we could discuss the wording on some labels further.

PR #1540, opened 2026-07-06
Replaced the fragile/wrong urlopen JSON mock

theresnotime:
- Un-nested test_create_user_and_editor to a real method of OAuthTestCase and removed the skipTest.
  - Replaced the fragile/wrong urlopen JSON mock with a patch of TWLight.users.models.editor_global_userinfo returning
  FAKE_GLOBAL_USERINFO — the idiomatic approach the rest of the suite uses (other tests pass global_userinfo explicitly). This
  guarantees the editor is saved before wp_editcount is read.
  - Removed the now-unused, misleading FAKE_IDENTITY_DATA fixture.
  - Added a comment explaining the save-before-read requirement, referencing T352896.

Assisted-by: Claude Opus 4.8
Bug: [T352896](https://phabricator.wikimedia.org/T352896)

### DIFF
--- a/TWLight/users/tests.py
+++ b/TWLight/users/tests.py
@@ -49,8 +49,6 @@ from TWLight.users.helpers.editor_data import (
     editor_make_block_dict,
 )
 
-FAKE_IDENTITY_DATA = {"query": {"userinfo": {"options": {"disablemail": 0}}}}
-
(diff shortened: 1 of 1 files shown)


suecarmol (approved):
This looks good, let's merge and see if it works in CICD

PR #1542, opened 2026-07-10
Add user agent to globaluserinfo call

theresnotime:
## Description
Set a compliant User-Agent header on globaluserinfo API requests

The _get_user_info_request helper in TWLight/users/helpers/editor_data.py issued requests via urllib.request.urlopen() without setting a User-Agent header, so requests went out with Python's default Python-urllib/x.y agent. Wikimedia blocks this generic agent per its User Agent policy, returning HTTP 403 Forbidden.

Add a descriptive `TWLIGHT_USER_AGENT` constant containing the application name, URL, and contact address (sourced from `TWLIGHT_ERROR_MAILTO`), and send it via urllib.request.Request on the globaluserinfo query.

## Phabricator Ticket
- https://phabricator.wikimedia.org/T431808

## How Has This Been Tested?
- Verified against the live endpoint from the running container: the globaluserinfo call now returns 200 with real data (`{'id': 69468694, 'name': 'TheresNoTime-WMF', 'editcount': 709}`) instead of 403 Forbidden, and the full OAuth login / editor-creation path completes without the 500.
- Ran the existing users test suite (python manage.py test TWLight.users.tests)

## Types of changes

### DIFF
--- a/TWLight/users/helpers/editor_data.py
+++ b/TWLight/users/helpers/editor_data.py
@@ -10,6 +10,13 @@ from djmail import template_mail
 
 logger = logging.getLogger(__name__)
 
+# Wikimedia's User-Agent policy requires a descriptive User-Agent with contact
+# information; the generic Python-urllib default is blocked with HTTP 403.
+# https://meta.wikimedia.org/wiki/User-Agent_policy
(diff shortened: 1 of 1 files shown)


DHardy-WMF (approved):
LGTM! Thanks for including fallbacks with those env vars as well.

PR #1549, opened 2026-08-17
Only mark reminder_email_sent when the renewal actually sends

theresnotime:
Add a diagnostic command for renewal emails (this can probably be dropped in the future if no longer needed)
Only mark reminder_email_sent when the renewal email _actually_ sends

Bug: [T407250](https://phabricator.wikimedia.org/T407250)


jsnshrmn (approved):
LGTM!

